PR #4, opened 2022-08-25
Rustfmt everything

oli-obk:


RalfJung:
Strange that there is such a big diff... Probably a different rustfmt config than Miri/rustc? Anyway, seems fine to me.

oli-obk:
The issue is that miri uses unstable rustfmt features. I did add what features I could on stable but the diff was even weirder

PR #5, opened 2022-08-26
Test that ui tests actually work

oli-obk:
also exposes where it's noisy and hard to actually ui test a crate itself


src/lib.rs
```diff
@@ -74,6 +74,13 @@ impl Default for Config {
     }
 }
 
+impl Config {
+    pub fn stderr_filter_dir(&mut self, path: &str, replacement: &'static str) {
+        self.stderr_filters
+            .push((Regex::new(path).unwrap(), replacement));
```
RalfJung, line 80: I am confused by this API. What does it have to do with directories? Why not just add it to `stderr_filters` (which is a public field)?

oli-obk: yea... I had an earlier version that took a `PathBuf` and canonicalized it, but I've fallen back to doing it at the caller site

PR #9, opened 2022-08-30
Allow ui_test tests to directly depend on the crate exposing said tests

oli-obk:
The only requirement on the miri side for this is to add `"build"` to the list of supported `cargo miri` subcommands. We'll also get a cheap way to add functions (e.g. exposing the magic extern functions) to all `*-dep` tests


Cargo.toml
```diff
@@ -1,26 +1,31 @@
...
 
 [dependencies]
 rustc_version = "0.4"
 colored = "2"
 # Features chosen to match those required by env_logger, to avoid rebuilds
-regex = { version = "1.5.5", default-features = false, features = ["perf", "std"] }
+regex = { version = "1.5.5", default-features = false, features = [
+    "perf",
+    "std",
+] }
```
RalfJung, line 20:
```suggestion
regex = {
    version = "1.5.5",
    default-features = false,
    features = ["perf", "std"],
}
```
Seems more readable to me? Same for `color-eyre` below.

src/dependencies.rs
```diff
@@ -29,9 +30,7 @@ pub fn build_dependencies(config: &Config) -> Result<Dependencies> {
     };
     let mut build = Command::new(program);
     build.args(args);
-    // HACK: we're using `cargo run` (or `cargo miri run`), because the latter does not
-    // support `cargo miri build` yet.
-    build.arg("run");
+    build.arg("build");
```
RalfJung, line 33: We could make this part of the `args` can the user can configure?

tests/integrations/basic-fail/tests/ui_tests.rs
```diff
@@ -6,7 +6,7 @@ fn main() -> ui_test::color_eyre::Result<()> {
     let mut config = Config {
         quiet: false,
         root_dir: "tests/actual_tests".into(),
-        dependencies_crate_manifest_path: Some("tests/dummy_crate/Cargo.toml".into()),
+        dependencies_crate_manifest_path: Some("Cargo.toml".into()),
         dependency_builder: Some(DependencyBuilder {
             program: std::path::PathBuf::from("cargo"),
             args: vec![],
```
RalfJung, line 12:
```suggestion
            args: vec!["build"],
```

Cargo.toml
oli-obk: that isn't valid unfortunately. I used `[dependencies.regex]` instead

tests/integrations/basic-fail/tests/ui_tests.rs
```diff
@@ -6,10 +6,10 @@ fn main() -> ui_test::color_eyre::Result<()> {
     let mut config = Config {
         quiet: false,
         root_dir: "tests/actual_tests".into(),
-        dependencies_crate_manifest_path: Some("tests/dummy_crate/Cargo.toml".into()),
+        dependencies_crate_manifest_path: Some("Cargo.toml".into()),
         dependency_builder: Some(DependencyBuilder {
             program: std::path::PathBuf::from("cargo"),
-            args: vec![],
+            args: vec!["build".into()],
             envs: vec![("CARGO_TARGET_DIR".into(), path.into())],
```
RalfJung, line 13: Should we set a suitable default dependency builder? I guess the only variable here is `path`, so either we find a way to not need that, or we add a function to compute a dependency builder given a target path.

oli-obk: we can compute it automatically by implementing a `Default` impl manually and removing the `Option` around it

oli-obk: > either we find a way to not need that,

it's not needed, just convenient to avoid rebuilds. I defaulted it to nothing and will keep using it to overwrite the `CARGO_TARGET_DIR`

PR #12, opened 2022-10-18
add repository and documentation tag to Cargo.toml

hellow554:
If you don't like the documentation tag, because it leads to the rust repo and maybe divert from your code or is not applicable, I can remove that :)


oli-obk:
Yea, I think we should leave out the doc link and let crates.io fill in the docs.rs link. The rustc ui tests are not compatible with this crate anymore

PR #17, opened 2022-10-24
replaced crossbeam with scoped threads

Dajamante:
Removed crossbeam, replaced with `thread::scope` and `crossbeam_channel`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,9 @@ use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::process::{Command, ExitStatus};
 use std::sync::atomic::{AtomicUsize, Ordering};
+use crossbeam_channel::unbounded;
 use std::sync::Mutex;
+use std::thread;
 
 pub use color_eyre;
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -172,18 +173,17 @@ pub fn run_tests_generic(config: Config, file_filter: impl Fn(&Path) -> bool + S
...
-
+    let (submit, receive) = unbounded();
     // Some statistics and failure reports.
     let failures = Mutex::new(vec![]);
     let succeeded = AtomicUsize::default();
     let ignored = AtomicUsize::default();
     let filtered = AtomicUsize::default();
 
-    crossbeam::scope(|s| -> Result<()> {
+    thread::scope(|s| -> Result<()> {
```
oli-obk, line 183: CI is failing because `thread` is not in scope

```suggestion
    std::thread::scope(|s| -> Result<()> {
```

Dajamante: Sorry, that's what I get for pushing in hurry and panic.
I amended the file!

Issue #15, opened 2022-10-22
'only-on-host' tests do not get run when the target is explicitly set to be the host


RalfJung:
The Miri test suite always sets the `target`, it just sometimes happens to be equal to the host. That makes it much easier to consistently manage printing the tested target to the log, and requires fewer special cases everywhere.

However it seems like ui_test does not run `only-on-host` tests when the target is explicitly set to the host. That's a problem since it means the extern-so tests in Miri are actually never run right now.

RalfJung:
Here's the line that implements this logic:

https://github.com/oli-obk/ui_test/blob/944438e6f95c06268d0c628184301af4be2b3dce/src/lib.rs#L697

This should also return `true` if `config.target` is equal to the host target of the current rustc.

PR #19, opened 2022-10-29
fix on-host condition for the case where the target is explicitly set to be the host

RalfJung:
The inner functions no expect target/host to be `Some`; `run_tests_generic` sets appropriate values if they start out as `None`.

Fixes https://github.com/oli-obk/ui_test/issues/15

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -117,6 +117,20 @@ impl Config {
         }
         Ok(())
     }
+
+    /// Make sure we have the host and target triples.
+    pub fn fill_host_and_target(&mut self) {
+        if self.host.is_none() {
+            self.host = Some(
+                rustc_version::VersionMeta::for_command(std::process::Command::new(&self.program))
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -168,9 +166,21 @@ pub fn run_file(mut config: Config, path: &Path) -> Result<std::process::ExitSta
...
+    if config.host.is_none() {
+        config.host = Some(
+            rustc_version::VersionMeta::for_command(std::process::Command::new(&config.program))
+                .expect("failed to parse rustc version info")
+                .host,
+        );
+    }
+    if config.target.is_none() {
+        config.target = Some(config.host.clone().unwrap());
+    }
```
oli-obk, line 183: I think this should be a method on `Config` to allow easier re-use in the code paths that don't use `run_tests`

RalfJung: Fair, done.

Issue #14, opened 2022-10-20
Cannot handle non-utf8 in stderr/stdout


oli-obk:
Attempting to do this causes something in compiletest to crash.

```
     Running tests/compiletest.rs (target/debug/deps/compiletest-5625b9d9b19cedc6)
## Running ui tests in tests/pass against miri for x86_64-unknown-linux-gnu
   Compiler flags: ["--error-format=json", "--edition", "2018", "-Dwarnings", "-Dunused", "-Zui-testing", "--target", "x86_64-unknown-linux-gnu"]
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 31, error_len: Some(1) }
Location: /home/cbiffle/.cargo/registry/src/github.com-1ecc6299db9ec823/ui_test-0.3.1/src/lib.rs:467
```

This is because the test framework does this (`ui_test/src/lib.rs` line 466):

```rust
    let diagnostics = rustc_stderr::process(path, stderr);
    let stdout = std::str::from_utf8(stdout).unwrap();  // <-- not good
```

So, evidently, tests are forbidden from printing anything in non-UTF-8. Miri, however, handles it fine, as attested by being able to `./miri run` that same test case.

_Originally posted by @cbiffle in https://github.com/rust-lang/miri/issues/2606#issuecomment-1284826843_

PR #21, opened 2022-10-29
don't panic on non-UTF-8 output

RalfJung:
This won't let us actually ensure that non-UTF-8 output exactly matches the reference (basically all non-UTF-8 is normalized through the 'lossy' functions), but it will at least not panic.

Helps with https://github.com/oli-obk/ui_test/issues/14

### DIFF
--- a/src/dependencies.rs
+++ b/src/dependencies.rs
@@ -54,8 +54,8 @@ pub fn build_dependencies(config: &Config) -> Result<Dependencies> {
     let output = build.output()?;
 
     if !output.status.success() {
-        let stdout = String::from_utf8(output.stdout)?;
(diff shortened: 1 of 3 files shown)


oli-obk:
Hmm... I was planning on making everything use Vec<u8> and only doing the conversion during rendering of the diff. I guess we could do this PR and keep the issue open for the more general version

RalfJung:
JSON parsing will also need strings I think.

But yeah this is mostly meant as a stopgap to not fail quite so badly when there is non-UTF-8 data.

RalfJung:
Closing in favor of https://github.com/oli-obk/ui_test/pull/23.

PR #23, opened 2022-10-31
Check non-utf8 output instead of panicking

### LINKED ISSUE #14: Cannot handle non-utf8 in stderr/stdout

oli-obk:
Alternative to https://github.com/oli-obk/ui_test/pull/21

fixes #14


src/diff.rs
```diff
@@ -152,10 +154,19 @@ impl<'a> DiffState<'a> {
...
+    if let (Ok(expected), Ok(actual)) = (std::str::from_utf8(expected), std::str::from_utf8(actual))
+    {
+        let mut state = DiffState::default();
+        for row in lines(expected, actual) {
+            state.row(row);
+        }
+        state.finish();
+    } else {
+        eprint!("{}", "-".red());
+        std::io::stderr().write(expected).unwrap();
```
RalfJung, line 167: This will print many lines, right? That would look rather strange I think.
```
-the_output
some non-utf-8
+the_output
other non-utf-8
```
Maybe it should just do the diff on the utf-8-lossy result but print a warning saying the diff might miss some information?

src/lib.rs
```diff
@@ -164,7 +173,10 @@ pub fn run_tests(mut config: Config) -> Result<()> {
 pub fn run_file(mut config: Config, path: &Path) -> Result<std::process::ExitStatus> {
     config.build_dependencies_and_link_them()?;
 
-    let comments = Comments::default();
+    let comments = match std::fs::read_to_string(path) {
+        Ok(content) => Comments::parse(path, &content)?,
+        Err(_) => Comments::default(),
```
RalfJung, line 178: This seems bad, if there is a single non-utf8-character we just entirely ignore all comments in that file?

src/rustc_stderr.rs
```diff
@@ -125,20 +122,16 @@ pub(crate) fn filter_annotations_from_rendered(rendered: &str) -> std::borrow::C
...
 pub(crate) fn process(file: &Path, stderr: &[u8]) -> Diagnostics {
-    let stderr = std::str::from_utf8(stderr).unwrap();
-    let mut rendered = String::new();
+    let mut rendered = Vec::new();
     let mut messages = vec![];
     let mut messages_from_unknown_file_or_line = vec![];
-    for line in stderr.lines() {
-        if line.starts_with('{') {
-            match serde_json::from_str::<RustcMessage>(line) {
+    for (line_number, line) in stderr.split_inclusive(|&c| c == b'\n').enumerate() {
```
RalfJung, line 128: `&[u8]` implements `BufRead` which has a `lines` method, can't we use that?

oli-obk: because that produces `String`s again

RalfJung: Ah, bummer.

This is not entirely identical though, `lines()` also has handling for `\r`.

src/lib.rs
oli-obk: I've decided to just error if a special comment has non-utf8 before the end of line

src/rustc_stderr.rs
oli-obk: I switched to bstr's `lines()` function which works correctly

src/diff.rs
```diff
@@ -152,9 +152,19 @@ impl<'a> DiffState<'a> {
...
 
-pub fn print_diff(expected: &str, actual: &str) {
+pub fn print_diff(expected: &[u8], actual: &[u8]) {
+    let expected_str = String::from_utf8_lossy(expected);
+    let actual_str = String::from_utf8_lossy(actual);
+
+    if expected_str.as_bytes() != expected || actual_str.as_bytes() != actual {
+        eprintln!(
+            "{}",
+            "Non-UTF8 characters in output, diff may be imprecise around the Non-UTF8 chars".red()
```
RalfJung, line 162:
```suggestion
            "Non-UTF8 characters in output, diff may be imprecise.".red()
```

RalfJung:
Yes, I like this. :-)

PR #26, opened 2022-11-07
Add semver checks to CI

oli-obk:
using https://github.com/obi1kenobi/cargo-semver-check to automatically fail CI if a change is done without the appropriate `Cargo.toml` version bump

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,8 +10,8 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  style:
-    name: style checks
+  fmt:
+    name: check rustfmt
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


RalfJung:
So the idea is that we have to do the bump already when the patch lands, instead of bumping later when we make the release?

oli-obk:
That's the hope, yea.

oli-obk:
but it seems to not work 😆 it picked 0.3.0 as the base for the comparison

oli-obk:
ah... it's checking whether the current bump is legal I think? So we merge this, then rerun CI on your PR and we'll see what happens.

RalfJung:
Shouldn't it fail currently since this repo contains semver-breaking changes compared to 0.3.1?

PR #27, opened 2022-11-07
test patch bump

oli-obk:


oli-obk:
according to the semver checks we can just do a minor bump 😬

RalfJung:
That's wrong though? We added new fields to `pub struct Config`, changed `pub type Filter`, ...

oli-obk:
I know... I'm just trying to figure out what's going on

Issue #29, opened 2022-11-20
`cargo test` fails and leaves some files changed


RalfJung:
Running `cargo test` in a checkout of this repo fails when I run it locally:
```
$ cargo test
   Compiling ui_test v0.4.0 (/home/r/src/rust/lib-crates/ui_test)
    Finished test [unoptimized + debuginfo] target(s) in 3.65s
     Running unittests src/lib.rs (target/debug/deps/ui_test-a70863804512cefe)

running 13 tests
test parser::tests::parse_missing_level ... ok
test parser::tests::parse_slash_slash_at ... ok
test parser::tests::missing_colon_fail ... ok
test parser::tests::parse_simple_comment ... ok
test parser::tests::parse_slash_slash_at_fail ... ok
test parser::tests::parse_x86_64 ... ok
test tests::issue_2156 ... ok
test tests::missing_implicit_warn_pattern ... ok
test tests::duplicate_pattern ... ok
test tests::missing_pattern ... ok
test tests::find_pattern ... ok
test tests::missing_warn_pattern ... ok
test parser::tests::parse_regex_error_pattern ... ok

test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/integration.rs (target/debug/deps/integration-dfceebd5da778812)

Running `integrations` tests in mode pass
...
test result: ok. 3 tests passed, 0 ignored, 0 filtered out


Running `integrations` tests in mode fail
F
tests/integrations/basic-fail/Cargo.toml FAILED:
command: "cargo" "test" "--color" "never" "--jobs" "1" "--target-dir" "target" "--manifest-path" "tests/integrations/basic-fail/Cargo.toml" "--" "--test-threads" "1"

actual output differed from expected
--- tests/integrations/basic-fail/Cargo.stderr
+++ <stderr output>
... 97 lines skipped ...
 
 test result: FAIL. 3 tests failed, 0 tests passed, 0 ignored, 0 filtered out
-error: test failed, to rerun pass '--test ui_tests'
+error: test failed, to rerun pass `--test ui_tests`
+
+Caused by:
+  process didn't exit successfully: `$DIR/target/debug/ui_tests-107eeeee069799c8 --test-threads 1` (exit status: 1)
 


full stderr:
   Compiling basic_fail v0.1.0 (/home/r/src/rust/lib-crates/ui_test/tests/integrations/basic-fail)
[...]

PR #30, opened 2022-11-20
improve error printed when the exit status is wrong

RalfJung:
There should probably be a ui test for this but due to https://github.com/oli-obk/ui_test/issues/29 I can't really add one.


tests/integrations/basic-fail/Cargo.stderr
```diff
@@ -37,6 +38,17 @@ For more information about this error, try `rustc --explain E0308`.
 
 
 
+tests/actual_tests/exit_code_fail.rs FAILED:
+command: "rustc" "--error-format=json" "--out-dir" "/tmp/.tmpp2eImK" "--edition=2021" "--extern" "basic_fail=$DIR/$DIR/../../../target/debug/libbasic_fail-$HASH.rmeta" "-L" "$DIR/$DIR/../../../target/debug" "-L" "$DIR/$DIR/../../../target/debug" "tests/actual_tests/exit_code_fail.rs"
+
+fail test got exit status: 0, but expected 1
+
+no error patterns found in failure test
```
RalfJung, line 46: So it's 'failure test' once and 'fail test' the other time. We should make that consistent.

PR #33, opened 2022-11-29
Revise revisions

### LINKED ISSUE #31: Support revisions on global commands
From what I could tell, revisions are currently only supported on error patterns. Would be nice to also support them on global commands, so that one can do things like
```
//@revisions: isolated free
//@only-target-linux
//[free]@compile-flags: -Zmiri-disable-isolation
```

FWIW I would have probably picked this syntax
```
//@revisions: isolated free
//@only-target-linux
//@[free]compile-flags: -Zmiri-disable-isolation
```
but that would be inconsistent with `//[fn_ptr]~^ ERROR: pattern`... maybe we should change that to `//~[fn_ptr]^ ERROR: pattern`, too, so that the first character after the `//` always indicates what this is?

OTOH, that might create parsing ambiguities if there is no ERROR level, and `//[revision]` is also consistent, so... whatever.

oli-obk:
fixes #31

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -682,14 +682,14 @@ fn check_annotations(
     let mut lowest_annotation_level = Level::Error;
     for &ErrorMatch {
         ref pattern,
-        revision: ref rev,
+        ref revisions,
         definition_line,
         line,
         level,
     } in &comments.error_matches
     {
-        if let Some(rev) = rev {
-            if rev != revision {
+        if !revisions.is_empty() {
+            if revisions.iter().all(|rev| rev != revision) {
                 continue;
             }
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -727,11 +727,21 @@ fn check_annotations(
     }
 
     let filter = |msgs: Vec<Message>, errors: &mut Vec<_>| -> Vec<_> {
+        let mut error = |_| {
+            errors.push(Error::InvalidComment {
+                msg: "`require_annotations_for_level` specified twice for same revision".into(),
+                line: 0,
+            })
+        };
         msgs.into_iter()
             .filter(|msg| {
                 msg.level
                     >= comments
-                        .find_one_for_revision(revision, |r| r.require_annotations_for_level, |_| errors.push(Error::InvalidComment { msg: "`require_annotations_for_level` specified twice for same revision".into(), line: 0 }))
+                        .find_one_for_revision(
+                            revision,
+                            |r| r.require_annotations_for_level,
+                            &mut error,
+                        )
                         .unwrap_or(lowest_annotation_level)
             })
             .collect()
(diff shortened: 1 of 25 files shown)


src/parser.rs
```diff
@@ -21,6 +21,40 @@ pub(crate) struct Comments {
     pub ignore: Vec<Condition>,
     /// Only run this test if all of these filters apply
     pub only: Vec<Condition>,
```
RalfJung, line 23: There's no fundamental reason these couldn't also be revisioned, is there?

src/parser.rs
```diff
@@ -21,6 +21,40 @@ pub(crate) struct Comments {
...
+    /// Comments that are only available under specific revisions.
+    /// The defaults are in key `vec![]`
+    pub revisioned: HashMap<Vec<String>, Revisioned>,
+}
+
+impl Comments {
+    pub fn find_one_for_revision<'a, T: 'a>(
```
RalfJung, line 30: This needs a doc comment. :)

src/parser.rs
oli-obk: Yea, I thought for a moment they couldn't be checked, but I think we have all the information available

src/lib.rs
```diff
@@ -715,18 +726,18 @@ fn check_annotations(
...
     }
 
-    let filter = |msgs: Vec<Message>| -> Vec<_> {
+    let filter = |msgs: Vec<Message>, errors: &mut Vec<_>| -> Vec<_> {
         msgs.into_iter()
             .filter(|msg| {
                 msg.level
                     >= comments
-                        .require_annotations_for_level
+                        .find_one_for_revision(revision, |r| r.require_annotations_for_level, |_| errors.push(Error::InvalidComment { msg: "`require_annotations_for_level` specified twice for same revision".into(), line: 0 }))
```
RalfJung, line 734: Can you let-bind the long `comments....` expression here? This is an awfully long line

RalfJung: Also I think this can be moved out of the `filter` loop?

oli-obk: oh wow rustfmt just completely gave up :D yea I'll clean that up

src/lib.rs
```diff
@@ -727,11 +727,21 @@ fn check_annotations(
...
+                msg: "`require_annotations_for_level` specified twice for same revision".into(),
+                line: 0,
+            })
+        };
         msgs.into_iter()
             .filter(|msg| {
                 msg.level
                     >= comments
-                        .find_one_for_revision(revision, |r| r.require_annotations_for_level, |_| errors.push(Error::InvalidComment { msg: "`require_annotations_for_level` specified twice for same revision".into(), line: 0 }))
+                        .find_one_for_revision(
```
RalfJung, line 740: Would it make sense to call this outside the `filter` closure? Right now we are re-doing this search over and over.

oli-obk: oh 🤦 I thought we needed the iteration

PR #35, opened 2022-12-13
Try running tests on multiple platforms

oli-obk:
work towards https://github.com/rust-lang/miri/issues/2724

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,6 +13,7 @@ use crossbeam_channel::unbounded;
 use parser::{ErrorMatch, Pattern, Revisioned};
 use regex::bytes::Regex;
 use rustc_stderr::{Diagnostics, Level, Message};
+use std::borrow::Cow;
 use std::collections::{HashSet, VecDeque};
 use std::ffi::OsString;
 use std::fmt::Display;
@@ -45,7 +46,8 @@ pub struct Config {
     pub host: Option<String>,
     /// `None` to run on the host, otherwise a target triple
     pub target: Option<String>,
-    /// Filters applied to stderr output before processing it
+    /// Filters applied to stderr output before processing it.
+    /// By default contains a filter for replacing backslashes with regular slashes.
     pub stderr_filters: Filter,
     /// Filters applied to stdout output before processing it
     pub stdout_filters: Filter,
@@ -74,7 +76,7 @@ impl Default for Config {
             trailing_args: vec![],
             host: None,
             target: None,
-            stderr_filters: vec![],
+            stderr_filters: vec![(Match::Exact(vec![b'\\']), b"/")],
             stdout_filters: vec![],
             root_dir: PathBuf::new(),
             mode: Mode::Fail {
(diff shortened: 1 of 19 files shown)


src/lib.rs
```diff
@@ -92,6 +92,25 @@ impl Default for Config {
...
+            .canonicalize()
+            .unwrap()
+            .parent()
+            .unwrap()
+            .display()
+            .to_string()
+            // Escape windows path backslashes so that regexes see the plain backslashes instead
+            // of escaping the follow-up symbol.
+            .replace('\\', r#"\\"#);
+        self.stderr_filter(pattern, replacement);
```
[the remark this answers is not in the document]
RalfJung: Oh I see, this is an even lower-level issue -- we are using regex normalization for the `$DIR` thing and that was going wrong.

Is there no function to turn a string into a regex that matches exactly that string, applying suitable escaping?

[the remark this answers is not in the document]
oli-obk: There is not, but I switched to just matching the vec exactly

src/lib.rs
```diff
@@ -169,7 +181,36 @@ pub enum OutputConflictHandling {
     Bless,
 }
 
-pub type Filter = Vec<(Regex, &'static [u8])>;
+/// A filter's match rule.
+#[derive(Debug)]
+pub enum Match {
+    /// If the regex matches, the filter applies
+    Regex(Regex),
+    /// If the path is found, the filter applies
```
RalfJung, line 189: "path"? There is no path here. Should probably be "if that exact sequence of bytes is found"?

src/lib.rs
```diff
@@ -74,7 +75,7 @@ impl Default for Config {
             trailing_args: vec![],
             host: None,
             target: None,
-            stderr_filters: vec![],
+            stderr_filters: vec![(Match::Exact(vec![b'\\']), b"/")],
```
RalfJung, line 78: Hard-coding a default filter is kind of suspicious. Why is this needed?

src/lib.rs
```diff
@@ -92,13 +93,24 @@ impl Default for Config {
 }
 
 impl Config {
+    /// Replace a path with a pattern if it appears in stderr.
```
RalfJung, line 96: It's replaced with `replacement`, not with a pattern, right?

```suggestion
    /// Add filter to replace given path if it appears in stderr.
```

src/lib.rs
oli-obk: ah yea, I had a `PathBuf` in there earlier

src/lib.rs
oli-obk: it's necessary, but not sufficient. This just gets rid of all the backslashes in stderr files without having to repeat that filter everywhere

RalfJung: I'm a bit worried about having hard-coded default filters without an opt-out.

oli-obk: we have an opt out, you can just overwrite the `stderr_filters` field, it's public.

RalfJung: Oh I see. Sounds good then.

RalfJung: But this should be documented somewhere.

oli-obk:
It works @RalfJung

PR #36, opened 2022-12-19
make quite not the default

RalfJung:
`cargo test` is not quiet per default, let's match that.


oli-obk:
I fixed clippy on main

PR #40, opened 2023-02-07
Add a new filter scheme for host triples

### LINKED ISSUE #38: Supporting only/ignore matching on the host
We have `only-target-foo` and `ignore-target-foo`, matching the target triple. It would make just as much sense to also have `only-host-foo` and `ignore-host-foo`, matching the host triple.

oli-obk:
fixes #38

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -102,7 +102,9 @@ impl<T> std::ops::DerefMut for CommentParser<T> {
 /// The conditions used for "ignore" and "only" filters.
 #[derive(Debug)]
 pub(crate) enum Condition {
-    /// The given string must appear in the target.
+    /// The given string must appear in the host triple.
+    Host(String),
+    /// The given string must appear in the target triple.
     Target(String),
     /// Tests that the bitwidth is the given one.
     Bitwidth(u8),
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -135,11 +137,13 @@ impl Condition {
...
-        } else if let Some(target) = c.strip_prefix("target-") {
-            Ok(Condition::Target(target.to_owned()))
+        } else if let Some(triple_substr) = c.strip_prefix("target-") {
+            Ok(Condition::Target(triple_substr.to_owned()))
+        } else if let Some(triple_substr) = c.strip_prefix("host-") {
+            Ok(Condition::Host(triple_substr.to_owned()))
         } else {
             Err(format!(
-                "invalid condition `{c:?}`, expected `on-host`, /[0-9]+bit/ or /target-.*/"
+                "invalid condition `{c:?}`, expected `on-host`, /[0-9]+bit/, /target-.*/, or /target-.*/"
```
RalfJung, line 146: Now it says `/target-.*/` twice...

oli-obk: Good job at copy paste oli... will fix tomorrow

PR #41, opened 2023-02-07
Add a manually run job for deploying to crates.io

oli-obk:


oli-obk:
@RalfJung deploying will now work by bumping the version in Cargo.toml and then running the release job manually on the Actions tab

oli-obk:
Could automate it to release when a version bump lands on the `main` branch

RalfJung:
> @RalfJung deploying will now work by bumping the version in Cargo.toml and then running the release job manually on the Actions tab

Wow, what is this witchcraft?^^ Personally I wouldn't entrust my registry token to GHA but sure...

I don't know what to press on the actions tab though. Or maybe this needs to land for a new button to appear or so?

oli-obk:
> Personally I wouldn't entrust my registry token to GHA but sure...

Ideally crates.io would support per crate tokens... but it doesn't, so...


Either way, you should be able to run the job at https://github.com/oli-obk/ui_test/actions/workflows/release.yml

Issue #58, opened 2023-04-28
GHA log folding causes conflicts when CI already has its own folding


RalfJung:
GHA log folding unfortunately does not nest. So ui-test emitting fold annotations clashes with Miri's own annotations, making the log a lot less readable.

Would be good if the folding annotations could be turned off.

oli-obk:
I still need to publish a version with this change, but then it becomes easy to fix on the miri side

PR #60, opened 2023-05-05
make gha emitting configurable

oli-obk:
allow miri to fix https://github.com/oli-obk/ui_test/issues/58 on its end.

### DIFF
--- a/src/github_actions.rs
+++ b/src/github_actions.rs
@@ -1,6 +1,6 @@
 //! An interface to github actions workflow commands.
 
-use std::fmt::Write;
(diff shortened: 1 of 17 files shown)


RalfJung:
I don't entirely follow what happens here. We probably *do* want the GHA "line annotations" for test failures in Miri, but we don't want the grouping. Is that possible or does this PR just let one choose all-or-nothing for GHA annotations?

oli-obk:
Yea I noticed that while working on the integration. I have a fix for that 😆

PR #61, opened 2023-05-05
Add env var to distinguish rustc aux vs test runs

bryangarza:
This patch just adds an env variable that scripts that wrap rustc can look at to know whether this is the actual test run itself, or just an auxilary call to rustc to gather some metadata.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -835,6 +835,7 @@ fn run_test(
         &mut errors,
     );
     cmd.args(&extra_args);
+    cmd.env("UITEST_TEST_RUN", "True");
 
     let output = cmd
         .output()
--- a/tests/integrations/basic/tests/ui_tests_with_wrapper_script.rs
+++ b/tests/integrations/basic/tests/ui_tests_with_wrapper_script.rs
@@ -1,0 +1,17 @@
+use std::num::NonZeroUsize;
+use ui_test::*;
+
+fn main() -> ui_test::color_eyre::Result<()> {
+    let config = Config {
+        quiet: false,
+        program: CommandBuilder::cmd("../runner.py"),
+        root_dir: "tests/actual_tests".into(),
+        // Never bless integrations-fail tests, we want to see stderr mismatches
+        output_conflict_handling: OutputConflictHandling::Error,
+        // Make sure our tests are ordered for reliable output.
+        num_test_threads: NonZeroUsize::new(1).unwrap(),
+        ..Config::default()
+    };
+
+    ui_test::run_tests(config)
+}
--- a/tests/integrations/basic-fail/Cargo.stderr
+++ b/tests/integrations/basic-fail/Cargo.stderr
@@ -7,7 +7,7 @@ tests/actual_tests/filters.rs ... FAILED
 tests/actual_tests/foomp.rs ... FAILED
 
 tests/actual_tests/bad_pattern.rs FAILED:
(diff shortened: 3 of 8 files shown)


src/lib.rs
```diff
@@ -1027,6 +1027,7 @@ fn run_test(
         &mut errors,
     );
     cmd.args(&extra_args);
+    cmd.env("RUSTC_UITEST_TEST_RUN", "True");
```
oli-obk, line 1030: I think we can leave off the `RUSTC` prefix, this crate can run a lot of other programs, too

oli-obk:
I'm thinking about how to best test this... I guess similar to https://github.com/oli-obk/ui_test/blob/main/tests/integrations/basic-fail/tests/ui_tests_invalid_program2.rs but as a regular run_tests call with a custom script. Could include tests for without aux build, with aux build, and sth where the aux build fails?

bryangarza:
> I'm thinking about how to best test this... I guess similar to https://github.com/oli-obk/ui_test/blob/main/tests/integrations/basic-fail/tests/ui_tests_invalid_program2.rs but as a regular run_tests call with a custom script. Could include tests for without aux build, with aux build, and sth where the aux build fails?

Sounds good, will add something

bryangarza:
@oli-obk Is the WIP commit is kind of what you had in mind (wasn't sure exactly what you meant by aux build but I basically added a wrapper script that just calls rustc)? But for some reason even when running `cargo test -- -- --bless` I can't get all the tests to pass :/ maybe you have some idea of what I'm doing wrong?
This is my test output:
[cargotestoutput.txt](https://github.com/oli-obk/ui_test/files/11560803/cargotestoutput.txt)

tests/integrations/basic/tests/ui_tests_with_wrapper_script.rs
```diff
@@ -0,0 +1,17 @@
+use std::num::NonZeroUsize;
+use ui_test::*;
+
+fn main() -> ui_test::color_eyre::Result<()> {
+    let config = Config {
+        quiet: false,
+        program: CommandBuilder::cmd("../runner.py"),
```
oli-obk, line 7: this makes you lose the default arguments (the ones that cause diagnostics to be rendered as json). Ideally just use the default rustc one and overwrite just the `config.program.program` part.

tests/integrations/basic/tests/ui_tests_with_wrapper_script.rs
```diff
@@ -0,0 +1,17 @@
...
+        root_dir: "tests/actual_tests".into(),
+        // Never bless integrations-fail tests, we want to see stderr mismatches
+        output_conflict_handling: OutputConflictHandling::Error,
+        // Make sure our tests are ordered for reliable output.
+        num_test_threads: NonZeroUsize::new(1).unwrap(),
+        ..Config::default()
+    };
+
+    ui_test::run_tests(config)
+}
```
oli-obk, line 17: All your new files seem to be missing the trailing line ending

oli-obk:
ping ^^

bryangarza:
> ping ^^

Thanks, yeah I had gotten stuck when trying to set up tests for this. When I get a chance I’ll push what I have to a separate branch, maybe you might have some guidance.

oli-obk:
I think this can be done by ui_test users that need it now that I have replaced the aux scheme with a pluggable system

PR #96, opened 2023-07-06
error when an error pattern comment refers to a line that doesn't exist

asquared31415:
A `-` without checks meant that determining what line a `//~^` was pointing to could overflow. Emit an error since that's never a well formed comment.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -596,7 +596,12 @@ impl CommentParser<&mut Revisioned> {
             ),
             Some('^') => {
                 let offset = pattern.chars().take_while(|&c| c == '^').count();
-                (self.line - offset, &pattern[offset..])
+                if let Some(match_line) = self.line.checked_sub(offset) {
+                    (match_line, &pattern[offset..])
+                } else {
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -596,7 +596,12 @@ impl CommentParser<&mut Revisioned> {
             ),
             Some('^') => {
                 let offset = pattern.chars().take_while(|&c| c == '^').count();
-                (self.line - offset, &pattern[offset..])
+                if let Some(match_line) = self.line.checked_sub(offset) {
+                    (match_line, &pattern[offset..])
+                } else {
+                    self.error(format!("//~^ pattern is trying to refer to {} lines previous, but is only on line {}", offset, self.line));
```
oli-obk, line 602:
```suggestion
                    self.error(format!("//~^ pattern is trying to refer to {} lines above, but the file only has {} lines", offset, self.line));
```

asquared31415: "above" is better, but the fact that it's the currently processed line matters, to me "the file only has {} lines" reads as "the whole file", even in context. how about "trying to refer to {} lines above, but there are only {} lines above it"?

oli-obk: Oh yea 🤦 that's much better

PR #97, opened 2023-07-12
Guard against crate id collisions

oli-obk:
see https://github.com/rust-lang/rust/issues/113585

### DIFF
--- a/src/dependencies.rs
+++ b/src/dependencies.rs
@@ -80,20 +80,24 @@ pub fn build_dependencies(config: &mut Config) -> Result<Dependencies> {
     let artifact_output = output.stdout;
     let artifact_output = String::from_utf8(artifact_output)?;
     let mut import_paths: HashSet<PathBuf> = HashSet::new();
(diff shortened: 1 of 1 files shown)


src/dependencies.rs
```diff
@@ -80,20 +80,24 @@ pub fn build_dependencies(config: &mut Config) -> Result<Dependencies> {
...
-            } else {
-                None
+    let mut artifacts = HashMap::new();
+    for line in artifact_output.lines() {
+        let Ok(message) = serde_json::from_str::<cargo_metadata::Message>(line) else {
+            continue
+        };
+        if let cargo_metadata::Message::CompilerArtifact(artifact) = message {
+            for filename in &artifact.filenames {
+                import_paths.insert(filename.parent().unwrap().into());
```
RalfJung, line 90: Should this assert that no existing entry is overwritten?

oli-obk: No, this doesn't matter, it is just about import paths, and these are the same for most dependencies.

PR #99, opened 2023-07-12
Stop comparing status code of compiler with mode if mode is Run

### LINKED ISSUE #98: executing binaries and expecting panics
Hey, so I am currently trying to use your crate to compile and check some binaries and I ran into a problem.
Basically, I have some files that I expect to panic when running them. So I annotated them with this comment:
`//@run: 101`
I just use the basic `Config::rustc()`.

However, this does not seem to work. It already complains after compilation, that the compilation did not exit with the expected error code (expected 101, actual 0).
I think (or hope) that this is not actually the intended behavior, right? If I specify an exit code with the `run` comment, I want to refer to the executed binary's exit code, not the one of the compiler.

I think the issue comes from these lines of code: [link](https://github.com/oli-obk/ui_test/blob/3b9edc5651af498fd5c27957c41ebbfa66e201f2/src/lib.rs#L535C1-L540C6).
You obtain the mode (and exit code) according to the file's comments, and check the compiler's exit code against it, as far as I understand it.
From my understanding, when the mode is `Run`, we should always expect the compiler to exit with code 0.

cedihegi:
Fixes #98


oli-obk:
Oh good catch. Please add a test for this in https://github.com/oli-obk/ui_test/tree/main/tests/integrations/basic-fail/tests/actual_tests_bless

cedihegi:
I added a very simple test for it.

oli-obk:
CI is failing, does `cargo test` work for you? I think it should modify some `Cargo.stderr` file. You then need to commit those changes

cedihegi:
aah sorry, yes locally it worked but I didn't commit all the changes it made.
I pushed it now, and this `revisioned` file referring to my test too. hope it works

oli-obk:
Thanks!

PR #101, opened 2023-07-13
Run rustfix by default if there are applicable suggestions

Alexendoo:
Runs rustfix by default on tests if there are applicable suggestions, unless the test has `//@no-rustfix` or there's `rustfix: false` in the config

This should be nicer in clippy tests than the old mechanism where we read a text file emitted by `compiletest-rs`, [storing the exceptions separately](https://github.com/rust-lang/rust-clippy/blob/7ccf5d404b0e044cfd1732aa5c33b16ad2c53d48/tests/compile-test.rs#L361-L398)


oli-obk:
I think we could just make "run rustfix if there are machine applicable suggestions" the default mode, remove the run rustfix flag and add your no rustfix flag

Alexendoo:
Tested that with clippy and yeah that works nicely. Added an opt out to the config as well so we don't try to parse e.g. cargo output

PR #102, opened 2023-07-14
Add a matcher for windows path backslashes

### LINKED ISSUE #89: Use a backslash escaping heuristic to keep non-path backslashes around
https://github.com/rust-lang/rust-clippy/pull/10426#discussion_r1179125971

> Could we adapt [the path heuristic from rust's compiletest](https://github.com/rust-lang/rust/blob/6ce22733b973355573efd1e6294e585460e90e17/src/tools/compiletest/src/runtest.rs#L3748-L3775) here? I think it would need support in ui_test as it uses a non string Replacer

> It would be nice to have non path backslashes stay as backslashes as that frequently confuses people writing tests

Alexendoo:
Fixes #89

Slightly different than [the rustc one](https://github.com/rust-lang/rust/blob/6ce22733b973355573efd1e6294e585460e90e17/src/tools/compiletest/src/runtest.rs#L3748-L3775), it assumes it will be ran against generally unsubstituted paths (no `$DIR\foo\...`) because `Config::path_stderr_filter` switches the `\`s in windows paths to `/`

It does require an extra feature enabled on `regex` for the unicode property escapes

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -43,8 +43,14 @@ jobs:
           - os: windows-latest
             host_target: i686-pc-windows-msvc
     runs-on: ${{ matrix.os }}
+    # Run tests under a directory with a space in it to double check the windows path heuristic
+    defaults:
+      run:
+        working-directory: "dir with spaces/ui test"
     steps:
     - uses: actions/checkout@v3
+      with:
+        path: "dir with spaces/ui test"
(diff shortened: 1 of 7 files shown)


oli-obk:
You need to run `cargo test` and commit the changed `.stderr` files

oli-obk:
cool, I can probably start removing a bunch of weird ad-hoc filters in my tests now :D

PR #121, opened 2023-07-29
Detect anything starting like a command, not just things that would parse

oli-obk:


oli-obk:
all of these are too prone to produce false alarms

PR #132, opened 2023-08-07
Avoid a few avenues for spurious failures

oli-obk:
see https://github.com/rust-lang/miri/pull/3008 for some debugging around this

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,8 +25,8 @@ use std::ffi::OsString;
 use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::process::{Command, ExitStatus, Stdio};
-use std::thread;
 use std::time::Duration;
+use std::{io, thread};
 
 use crate::parser::{Comments, Condition};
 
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -649,6 +648,21 @@ impl dyn TestStatus {
     }
 }
 
+fn retry_with_backoff<T>(mut f: impl FnMut() -> io::Result<T>) -> io::Result<T> {
+    let mut res = f();
+    for i in 0..10 {
+        if let Err(err) = &res {
+            if let io::ErrorKind::WouldBlock = err.kind() {
+                std::thread::sleep(Duration::from_millis(100 * i));
```
RalfJung, line 656:
```suggestion
                std::thread::sleep(Duration::from_millis(100 * i + 10));
```
Let's sleep a bit also after the very first attempt?

oli-obk: oh 😆 good point

src/lib.rs
```diff
@@ -649,6 +648,21 @@ impl dyn TestStatus {
     }
 }
 
+fn retry_with_backoff<T>(mut f: impl FnMut() -> io::Result<T>) -> io::Result<T> {
+    let mut res = f();
+    for i in 0..10 {
+        if let Err(err) = &res {
+            if io::ErrorKind::WouldBlock == err.kind() || i < 5 {
```
RalfJung, line 655: Do you really want to retry on arbitrary failures? That sounds wrong to me.

oli-obk: Well... We also get arbitrary permission denied failures on process spawn, and I have no clue why

Issue #135, opened 2023-08-09
ui_test got slower


oli-obk:
Nice, runs fine now

I've noticed it's quite a bit slower, measuring it with:

```bash
hyperfine -L commit 7c595b4599d546dc5f36bc4bebbbf8ad454fed62,ui_test \
    --prepare 'git checkout {commit}' \
    --warmup 2 \
    -n '{commit}' \
    'cargo uitest'
```

x86_64-pc-windows-msvc, 6 cores/12 threads

```
Benchmark 1: 7c595b4599d546dc5f36bc4bebbbf8ad454fed62
  Time (mean ± σ):     18.325 s ±  0.296 s    [User: 113.464 s, System: 59.451 s]
  Range (min … max):   17.899 s … 18.837 s    10 runs

Benchmark 2: ui_test
  Time (mean ± σ):     24.262 s ±  0.041 s    [User: 61.322 s, System: 39.990 s]
  Range (min … max):   24.208 s … 24.332 s    10 runs

Summary
  '7c595b4599d546dc5f36bc4bebbbf8ad454fed62' ran
    1.32 ± 0.02 times faster than 'ui_test'
```

aarch64-unknown-linux-gnu, 4 cores/4 threads

```
Benchmark 1: 7c595b4599d546dc5f36bc4bebbbf8ad454fed62
  Time (mean ± σ):     55.910 s ±  0.233 s    [User: 146.823 s, System: 58.316 s]
  Range (min … max):   55.658 s … 56.372 s    10 runs
 
Benchmark 2: ui_test
  Time (mean ± σ):     85.066 s ±  0.091 s    [User: 97.486 s, System: 48.101 s]
  Range (min … max):   84.942 s … 85.243 s    10 runs
 
Summary
  '7c595b4599d546dc5f36bc4bebbbf8ad454fed62' ran
    1.52 ± 0.01 times faster than 'ui_test'
```

_Originally posted by @Alexendoo in https://github.com/rust-lang/rust-clippy/issues/11239#issuecomment-1671221649_

PR #137, opened 2023-08-10
Various smaller fixes and perf improvements

oli-obk:


oli-obk:
possibly fixed https://github.com/oli-obk/ui_test/issues/135 by avoiding the extra clones introduced in https://github.com/oli-obk/ui_test/commit/c0f809e877b7cb731c7a1a4b8954ac84e7eb507c

This was the only regression noticeable in this repo. We'll have to investigate clippy again after this is released

PR #148, opened 2023-08-25
Allow configuring `rustfix::Filter` independently of the mode

Alexendoo:
Since https://github.com/rust-lang/rust-clippy/pull/11249 was merged we'll be wanting to move clippy from `Yolo` to `Fail` to make the annotations required, but we still want to apply non machine applicable suggestions so this adds an enum to configure rustfix

cc @GuillaumeGomez since https://github.com/rust-lang/rust-clippy/pull/11249#issuecomment-1688502203 will be blocked on this

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,6 +1,6 @@
 use regex::bytes::Regex;
 
-use crate::{dependencies::build_dependencies, CommandBuilder, Filter, Match, Mode};
+use crate::{dependencies::build_dependencies, CommandBuilder, Filter, Match, Mode, RustfixMode};
 pub use color_eyre;
 use color_eyre::eyre::Result;
 use std::{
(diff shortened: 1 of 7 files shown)


oli-obk:
Needs rustfmt, otherwise lgtm.
We can remove Yolo mode now, too, but that'll require some test suite changes. I only added Yolo mode for clippy, we don't need it anymore

Alexendoo:
Ahh I was still on 1.71

Looks like [MIRI uses Yolo in some form](https://github.com/rust-lang/miri/blob/0749bb55e8fe2d3ec4f3c73c9b7255d88f13f532/tests/compiletest.rs#L288), not sure on the context how necessary it is though

oli-obk:
I can publish on Monday, or, if you open a PR with version bumps, I can tell GHA to publish

PR #156, opened 2023-08-30
Fast rerun

### LINKED ISSUE #143: Don't rerun tests if nothing changed
compiletest is able to cache test results. Would be great if ui_test could also do that. :)

oli-obk:
fixes #143

cc @RalfJung

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,8 +148,8 @@ pub fn default_file_filter(path: &Path, config: &Config) -> bool {
 ///
 /// To only include rust files see [`default_file_filter`].
 pub fn default_any_file_filter(path: &Path, config: &Config) -> bool {
-    let path = path.display().to_string();
-    let contains_path = |files: &[String]| files.iter().any(|f| path.contains(f));
+    let path_str = path.display().to_string();
+    let contains_path = |files: &[String]| files.iter().any(|f| path_str.contains(f));
 
     if contains_path(&config.skip_files) {
         return false;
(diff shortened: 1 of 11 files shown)


RalfJung:
So what's the logic here -- ui_test keeps track of a timestamp file, and if a test is older than that file it gets rerun? Also if the test binary changes everything gets rerun? I guess that works for Miri since the compiletest binary actually links against Miri even though it doesn't have to... not sure if it will work in other cases. But I also don't have a better idea.

If a test failed last time, then rustc compiletest always reruns it.  That seems preferable since otherwise, the test suite can seem to pass (thanks to ignoring all failed tests) when it shouldn't pass. That logic is currently missing here, or did I misunderstand the code?

src/lib.rs
```diff
@@ -636,6 +661,13 @@ impl dyn TestStatus {
         run_rustfix(
             &stderr, &stdout, path, comments, revision, config, *mode, extra_args,
         )?;
+
+        // Make sure the `default_file_filter` recognizes that the test has already been run successfully
+        // and avoid running it again.
+        let check_file = config.out_dir.join(path.with_extension("timestamp"));
+        std::fs::create_dir_all(check_file.parent().unwrap()).unwrap();
+        std::fs::File::create(check_file).unwrap();
```
oli-obk, line 669: > If a test failed last time, then rustc compiletest always reruns it. 

This only happens here, but I realize now that we have an early success path for `run` tests 🙃 so `run` tests are always run again

oli-obk:
hmm... I just realized that this won't detect if an aux-build file changes 🙃

RalfJung:
Ah, makes sense, thanks! LGTM, though I'll only actually be sure when this gets tested for real...^^

> I just realized that this won't detect if an aux-build file changes 🙃

Ah, and that.^^

Alexendoo:
There's some tricky cases in clippy like the `clippy.toml` file changing for ui toml tests, and

```rust
#[path = auxiliary/foo.rs]
mod foo;
```
```rust
include_str("foo.txt");
```

This all should be contained in the depinfo file but handling that is a jump in complexity

RalfJung:
Ah yeah, Miri also uses something like that.

PR #157, opened 2023-08-30
Populate `Config` with `Args` instead of passing it to `run_tests_generic`

Alexendoo:
`Args` now solely parses arguments and is used to fill in fields on `Config`, the default number of test threads is also moved into `run_tests_generic`

This reintroduces `OutputConflictHandling` as the command string in the error variant is nice to have

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,6 +5,7 @@ pub use color_eyre;
 use color_eyre::eyre::Result;
 use std::{
     ffi::OsString,
+    num::NonZeroUsize,
     path::{Path, PathBuf},
 };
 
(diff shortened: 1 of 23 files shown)


src/config.rs
```diff
@@ -98,6 +111,42 @@ impl Config {
...
+    ///
+    /// The default output conflict handling command suggests adding `--bless`
+    /// to the end of the current command.
+    pub fn with_args(&mut self, args: &Args, default_bless: bool) {
+        let Args {
+            ref filters,
+            quiet: _,
+            check,
+            bless,
+            threads: _,
```
oli-obk, line 125: you added a threads field to `Config` but didn't fill it in here (in case it is `Some` here)

oli-obk:
Please also do a major version bump in this PR.

src/config.rs
Alexendoo: Whoops, I even did a destructure to avoid exactly that 😅

PR #161, opened 2023-09-05
Nextest

### LINKED ISSUE #158: Support flags needed for `cargo nextest`
Support for `--exact` and `--nocapture` would allow using `cargo nextest run`, currently it results in a panic

```
called `Result::unwrap()` on an `Err` value: unknown command line flag `--exact`: ["compile_test", "--nocapture"]
```

oli-obk:
fixes #158

cc @Alexendoo

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,7 +148,15 @@ pub fn default_file_filter(path: &Path, config: &Config) -> bool {
 /// To only include rust files see [`default_file_filter`].
 pub fn default_any_file_filter(path: &Path, config: &Config) -> bool {
     let path = path.display().to_string();
-    let contains_path = |files: &[String]| files.iter().any(|f| path.contains(f));
+    let contains_path = |files: &[String]| {
+        files.iter().any(|f| {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -229,6 +236,23 @@ pub fn run_tests_generic(
...
+    if configs.iter().any(|c| c.list) {
+        if configs.iter().any(|c| !c.run_only_ignored) {
+            println!("ui_test: test");
+        }
+        return Ok(());
+    }
+    for config in &mut configs {
+        if config.filter_exact
+            && config.filter_files.len() == 1
+            && config.filter_files[0] == "ui_test"
```
Alexendoo, line 249: If `TESTNAME` is used there could be some other filters in here since args extends the config's list

oli-obk: I guess... filtering out just one specific name seems not great tho

Alexendoo: Yeah I wouldn't be too fussed if you ignored that

PR #165, opened 2023-09-23
Allow patternless error and warning matchers

### LINKED ISSUE #164: //~ ERROR annotations without error body
On [this Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Improve.20UI.20error.20checks), we concluded that Clippy doesn't need bodies for the `//~ ERROR` annotations. It would be really cool if there wasn't a need to put the body of the error on the annotation (maybe with a `//@ optional-bodies` header or similar?)

Jarcho:
fixes #164

This adds support for patternless error matchers (//~ ERROR` and `//~ WARN`). These come with two restrictions:
* They can't be used on the same line as a matcher with a pattern
* All the patternless matchers in a single test have to be for the same error code.

The reason for the second restriction is the same general reason for having inline annotations in the first place. Since the `.stderror` file already has the full message the annotation doesn't test anything new, it only helps with reading/writing the test file. If only a single error is being tested, having a message with each annotation is unlikely to add anything but line noise. If there are multiple errors being tested for, then the message helps which error is expected for each annotation.


Jarcho:
The error message when an unexpected error code is still in need of improvement. I think merging it with the unmatched error message would be easier to read. Also noting which error caused the selection of the expected error code would be good to have.

RalfJung:
>  Since the .stderror file already has the full message the annotation doesn't test anything new, it only helps with reading/writing the test file.

I very strongly disagree. The stderr file is easily adjusted by re-blessing. The explicit annotations in the file are absolutely crucial to ensure that this test keeps testing the right thing even as things get re-blessed.

Of course your use of this crate may be different, but IMO relying just on stderr for the output message is way too fragile. I certainly regularly rely on compiletest / ui_test checking that the right message is emitted even in bless mode. I need bless mode because some details change, and it's good to see what changes, but I'd never want to bless changes to the relevant part of the message -- and it's way to easy to miss such changes when re-blessing everything for a large change, so IMO it is crucial to have the key parts of the error message in the rs test file.

oli-obk:
I agree with Ralf, but I also want to support clippy's use case. I think we should change the boolean in https://github.com/oli-obk/ui_test/blob/3d262a2be8cd2024c5108fa02d6f8047c4d61a75/src/mode.rs#L40 to be an enum, (`Require`, `Optional`, `RequireWithoutMessage`... the names aren't great, but obvious, and can get documented individually).

Jarcho:
This could also require the code to specified (e.g. `//@patternless-error-code:code`) which opts into this on each test. It also acts as a guard in case something crazy changes.

> Anyway, ui_test can of course have such a mode, but the downsides should be clearly documented (and ideally I could make sure that the Miri test suite never has any patternless matchers).

It looks like Miri doesn't fill in a code on any of it's diagnostics so the current implementation intentionally wouldn't match any of them.

Alexendoo:
How about in addition to e.g. `//~ ERROR: message` we accept `//~ clippy::some_lint`? It's shorter than the original `//~ ERROR clippy::some_lint` proposal. If dropping the `clippy::` prefix is allowed it's not that far away from `//~ ERROR` in length, aside from those few lints with hefty names

It's clear at the annotation site what is being expected, rather than having to specify/know the file context

It doesn't include the level but I wouldn't expect a test where the level of a single code varies in the same file to exist outside of rustc

Jarcho:
Implemented @Alexendoo's suggestion. This allows `//~ error_code` to match either a single warning or error with that error code.

The readme has also been reformatted slightly to mention both forms. The erroneous note about being able to drop the level when giving a pattern has been removed as well. The parser didn't accept that in the first place.

tests/integrations/basic-fail/tests/actual_tests/wrong_diagnostic_code.rs
```diff
@@ -0,0 +1,9 @@
+#![warn(dead_code, unreachable_code)]
```
oli-obk, line 1: Why is this test in basic-fail, and not somewhere where it'll get blessed?

src/lib.rs
```diff
@@ -1092,35 +1092,58 @@ fn check_annotations(
...
 
-        errors.push(Error::PatternNotFound {
-            pattern: pattern.clone(),
-            expected_line: Some(line),
+        errors.push(match kind {
+            ErrorMatchKind::Pattern { pattern, .. } => Error::PatternNotFound {
+                pattern: pattern.clone(),
+                expected_line: Some(line),
+            },
+            ErrorMatchKind::Code(code) => Error::CodeNotFound {
```
oli-obk, line 1143: Please add a test excercising this code path

src/lib.rs
```diff
@@ -1092,35 +1092,58 @@ fn check_annotations(
...
+                        .iter()
+                        .position(|msg| pattern.matches(&msg.message) && msg.level == level);
+                    if let Some(found) = found {
+                        msgs.remove(found);
+                        continue;
+                    }
+                }
+                ErrorMatchKind::Code(code) => {
+                    let found = msgs.iter().position(|msg| {
+                        msg.level >= Level::Warn
```
oli-obk, line 1127: Do you really want these tests to work on warnings? Could require all tests to deny the specific lints they care about

src/parser.rs
```diff
@@ -750,43 +764,52 @@ impl CommentParser<&mut Revisioned> {
...
+            let pattern = self.parse_error_pattern(pattern);
 
-        let pattern = self.parse_error_pattern(pattern);
+            self.error_matches.push(ErrorMatch {
+                kind: ErrorMatchKind::Pattern { pattern, level },
+                line: match_line,
+            });
+        } else if (*level_or_code).parse::<Level>().is_ok() {
+            // Shouldn't conflict with any real diagnostic code
+            self.error(pattern.span(), "no `:` after level found");
```
oli-obk, line 794:
```suggestion
            self.error(level_or_code.span(), "no `:` after level found");
```

oli-obk (changes requested):

tests/integrations/basic-fail/tests/actual_tests/wrong_diagnostic_code.rs
Jarcho: Is that not the correct spot for a test with an unmatched comment?

src/lib.rs
Jarcho: I don't see there being a problem either way. Only working with errors would force tests to be a little more consistent.

tests/integrations/basic-fail/tests/actual_tests/wrong_diagnostic_code.rs
oli-obk: I did not realize that's what you're testing. Please add a comment and move it to one of the bless directories. Also I don't think you're testing that from the diff I'm seeing, as it just looks like the test is missing its stderr output and is testing that a pass test in a fail directory will emit a message about the wrong exit code

src/lib.rs
oli-obk: OK, let's do that then

tests/integrations/basic-fail/tests/actual_tests/wrong_diagnostic_code.rs
Jarcho: The first comment has the wrong code. The correct code for the first one is `dead_code`. The second `unreachable_code` comment is correct.

I'll add a comment to the test.

oli-obk: Oh... maybe I'm not emitting annotation mismatch errors if there were stderr differences

Jarcho: I can try to move it and see what happens.

Jarcho: Moved it to `actual_tests_bless` which does give a `.stderr` file. Is this what you meant in the first place?

Jarcho:
Last commit should fix those comments. I'll squash it down if there aren't any issues.

tests/integrations/basic-fail/Cargo.stdout
```diff
@@ -755,6 +756,57 @@ full stderr:
...
+4 |     //~^ should_be_dead_code
+  |          ^^^^^^^^^^^^^^^^^^^ expected because of this pattern
+  |
+
+error: there were 1 unmatched diagnostics
+ --> tests/actual_tests_bless/wrong_diagnostic_code.rs:3:4
+  |
+3 | fn foo() -> i32 {
+  |    ^^^ Error: function `foo` is never used
+  |
```
oli-obk, line 775: Print the diagnostic code of unmatched diagnostics, otherwise it is very hard to figure out how to resolve this.

oli-obk (changes requested):

README.md
```diff
@@ -11,16 +11,16 @@ A smaller version of compiletest-rs
...
-    be careful not to accidentally have a pattern that differs between platforms.
-* if `XXX` is of the form `/XXX/` it is treated as a regex instead of a substring and will succeed if the regex matches.
+* `//~ LEVEL: XXX` matches by error level and message text
+    * `LEVEL` can be one of the following (descending order): `ERROR`, `HELP`, `WARN` or `NOTE`
+    * If a level is specified explicitly, *all* diagnostics of that level or higher need an annotation. To avoid this see `//@require-annotations-for-level`
+    * This checks the output *before* normalization, so you can check things that get normalized away, but need to
+        be careful not to accidentally have a pattern that differs between platforms.
+    * if `XXX` is of the form `/XXX/` it is treated as a regex instead of a substring and will succeed if the regex matches.
+* `//~ CODE` matches by diagnostic code.
+    * `CODE` can take multiple forms such as: `E####`, `lint_name`, `tool::lint_name`.
```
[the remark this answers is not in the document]
Jarcho: The failure is noted in the paragraph above introducing `//~` comments. Second note has been added.

tests/integrations/basic-fail/Cargo.stdout
Jarcho: It's been added only for errors. Should be added for warnings as well?

Jarcho:
Last change adds the ability to specify a global prefix when matching error codes. This is useful for clippy since all of ours start with `clippy::` and we don't need to match anything else.

src/config.rs
```diff
@@ -59,6 +59,9 @@ pub struct Config {
     pub run_only_ignored: bool,
     /// Filters must match exactly instead of just checking for substrings.
     pub filter_exact: bool,
+    /// Prefix added to all diagnostic code matchers. Note this will make it impossible
+    /// match codes which do not contain this prefix.
+    pub diagnostic_code_prefix: Option<String>,
```
oli-obk, line 64: if we are adding a config option anyway, we may as well go all the way and have a mode for `//~ LEVEL` and one for `//~ CODE`, and only permit the one selected.

oli-obk (changes requested):

tests/integrations/basic-fail/Cargo.stdout
oli-obk: the `//~ CODE` scheme does not work for warnings, so no.

src/config.rs
Alexendoo: I believe we a few tests that have rustc diagnostics as well as clippy ones to check for overlaps, so they would need both kinds

src/lib.rs
```diff
@@ -1010,6 +1010,7 @@ fn check_test_result(
         config,
         revision,
         comments,
+        config.diagnostic_code_prefix.as_deref(),
```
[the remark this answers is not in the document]
Jarcho: Somehow I missed that when looking at the argument list.

src/config.rs
Jarcho: We might also have a few lints which would benefit from having the messages checked as well.

Jarcho:
Build failures are caused by the upgrade to rust v1.73.

oli-obk (dismissed):

PR #167, opened 2023-09-25
make individual errors in a failed test more clearly visible

RalfJung:
In large errors I feel it is easy to miss when somewhere at the top it also says "got exit status: 0, but expected 1", or so.

Ideally all errors would do this consistently, e.g. start with a underlined title and then perhaps have more details. However that doesn't really work for the errors that show a span... but maybe ut's worth doing that at least for these two otherwise-easy-to-miss errors?

### DIFF
--- a/src/status_emitter.rs
+++ b/src/status_emitter.rs
@@ -201,11 +201,12 @@ struct TextTest {
 }
 
 impl TextTest {
+    /// Prints the user-visible name for this test.
     fn msg(&self) -> String {
         if self.revision.is_empty() {
             self.path.display().to_string()
         } else {
-            format!("{} ({})", self.path.display(), self.revision)
+            format!("{} (revision `{}`)", self.path.display(), self.revision)
         }
     }
 }
@@ -243,11 +244,7 @@ impl TestStatus for TextTest {
         stderr: &'a [u8],
         stdout: &'a [u8],
     ) -> Box<dyn Debug + 'a> {
-        let mut path = self.path.display().to_string();
-        if !self.revision.is_empty() {
-            write!(path, " (revision `{}`)", self.revision).unwrap();
-        }
-        let text = format!("{} {path}", "FAILED TEST:".bright_red());
+        let text = format!("{} {}", "FAILED TEST:".bright_red(), self.msg());
 
         println!();
         println!("{}", text.bold().underline());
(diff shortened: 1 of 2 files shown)


oli-obk:
We could also use the snippets logic we use for spanned errors, so you'd get it in the same format and style (red `error` followed by the message)

RalfJung:
Yeah that would be the other option I guess. We need to be careful to make this still clearly "smaller" than the separator that introduces the messages for each test failure:
```
tests/fail/function_pointers/abi_mismatch_simple.rs FAILED:
```
Maybe if it looks like this that is enough?
```
<underline><bold><red>FAILED TEST:</red> tests/fail/function_pointers/abi_mismatch_simple.rs</bold></underline>

<bold><red>error:</red> bad exit status</bold>

<bold><red>error:</red> substring not found</bold>
{span printed}

<bold><red>error:</red> unmatched diagnostics</bold>
{span printed}
```

oli-obk:
Yea that sounds good to me.

RalfJung:
This looks better I think, but it's not entirely the same tone of red somehow...

image

RalfJung:
Though rustc and cargo are not consistent either.^^

image


Anyway I made it all `bright_red`, now it is consistent within ui_test. I left `diff` at `red` though, not sure which one `git diff` uses.

RalfJung:
I think `git diff` uses the darker red.

src/status_emitter.rs
```diff
@@ -243,16 +243,14 @@ impl TestStatus for TextTest {
         stderr: &'a [u8],
         stdout: &'a [u8],
     ) -> Box<dyn Debug + 'a> {
+        let mut path = self.path.display().to_string();
+        if !self.revision.is_empty() {
+            write!(path, " (revision `{}`)", self.revision).unwrap();
+        }
```
RalfJung, line 249: Isn't there a function that returns this name-plus-revision? It's also printed while the tests are running, right? We should have that logic only once to keep it consistent...

RalfJung: Ah, found that function. It wasn't consistent, so the output changed a bit.

RalfJung:
Updated screenshot:

image

oli-obk:
I love it, thanks!

RalfJung:
Yeah I'm quite happy with it. Can't wait to see it in Miri. Just needs a release now. *nudge wink* ;)

oli-obk:
I can do releases from gha, I just can't do the version bump from it 😃

I'll do both tomorrow

RalfJung:
No worries, not urgent at all. :)

PR #169, opened 2023-09-25
be more specific about what the pattern expects

RalfJung:
In particular the current text is confusing if the substring *does* appear in stderr, but not on the right line.

### DIFF
--- a/tests/integrations/basic-fail/Cargo.stdout
+++ b/tests/integrations/basic-fail/Cargo.stdout
@@ -17,7 +17,7 @@ tests/actual_tests/rustc_ice.rs ... FAILED
 tests/actual_tests/bad_pattern.rs FAILED:
 command: "rustc" "--error-format=json" "--extern" "basic_fail=$DIR/$DIR/../../../target/$TMP/$TRIPLE/debug/libbasic_fail.rlib" "--extern" "basic_fail=$DIR/$DIR/../../../target/$TMP/$TRIPLE/debug/libbasic_fail-$HASH.rmeta" "-L" "$DIR/$DIR/../../../target/$TMP/$TRIPLE/debug" "-L" "$DIR/$DIR/../../../target/$TMP/$TRIPLE/debug" "--out-dir" "$TMP "tests/actual_tests/bad_pattern.rs" "--edition" "2021"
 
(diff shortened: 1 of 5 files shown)


oli-obk:
> the substring _does_ appear in stderr, but not on the right line.

I plan on creating a new diagnostic for this specific case, as it is very common. Basically merge the "missing annotation" and "annotation without error" messages and emit a single message

oli-obk:
oh right lol. that used to be how it worked in the very first version XD

tests/integrations/basic-fail/Cargo.stdout
```diff
@@ -17,7 +17,7 @@ tests/actual_tests/rustc_ice.rs ... FAILED
 tests/actual_tests/bad_pattern.rs FAILED:
 command: "rustc" "--error-format=json" "--extern" "basic_fail=$DIR/$DIR/../../../target/$TMP/$TRIPLE/debug/libbasic_fail.rlib" "--extern" "basic_fail=$DIR/$DIR/../../../target/$TMP/$TRIPLE/debug/libbasic_fail-$HASH.rmeta" "-L" "$DIR/$DIR/../../../target/$TMP/$TRIPLE/debug" "-L" "$DIR/$DIR/../../../target/$TMP/$TRIPLE/debug" "--out-dir" "$TMP "tests/actual_tests/bad_pattern.rs" "--edition" "2021"
 
-error: substring `miesmätsched types` not found in stderr output
+error: `miesmätsched types` not found in diagnostics for line 5
```
oli-obk, line 20: were you trying to refer to the line where the error occurs? because that's line 4. This information is currently not available where you're reporting the error.

RalfJung: I was referring to the line where the error is expected to occur. That is line 5, right? The error was "not found in line 5". Saying it was "not found in line 4" would be wrong since as you say it *was* found in line 4.

oli-obk: so... the annotation is in line 5, but it has a `~^`, so it refers to line 4. But the message was not found in line 4.

RalfJung: I think I fixed this.

Issue #168, opened 2023-09-25
Running `cargo test` leaves behind a diff


RalfJung:
After running `cargo test`, `git diff` shows:
```diff
diff --git a/tests/integrations/basic-fail/Cargo.stdout b/tests/integrations/basic-fail/Cargo.stdout
index 7b13acd..68ec71d 100644
--- a/tests/integrations/basic-fail/Cargo.stdout
+++ b/tests/integrations/basic-fail/Cargo.stdout
@@ -234,6 +234,24 @@ command: "rustc" "--error-format=json" "--extern" "basic_fail=$DIR/$DIR/../../..
 
 fail test got exit status: 101, but expected 1
 
+actual output differed from expected
+Execute `DO NOT BLESS. These are meant to fail` to update `tests/actual_tests/rustc_ice.stderr` to the actual output
+--- tests/actual_tests/rustc_ice.stderr
++++ <stderr output>
+ error: internal compiler error: no errors reported for args
+   --> $DIR/rustc_ice.rs:10:5
+... 4 lines skipped ...
+ thread 'rustc' panicked at 'aborting due to `-Z treat-err-as-bug=1`', compiler/rustc_errors/src/lib.rs
+ stack backtrace:
++                               at ./nptl/pthread_create.c:444:8
++                               at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
+ 
+ error: the compiler unexpectedly panicked. this is a bug.
+... 10 lines skipped ...
+ #2 [analysis] running analysis passes on this crate
+ end of query stack
+
+
 error: substring `mismatched types` not found in stderr output
   --> tests/actual_tests/rustc_ice.rs:11:17
    |
@@ -257,6 +275,8 @@ error: internal compiler error: no errors reported for args
 
 thread 'rustc' panicked at 'aborting due to `-Z treat-err-as-bug=1`', compiler/rustc_errors/src/lib.rs
 stack backtrace:
+                               at ./nptl/pthread_create.c:444:8
+                               at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
 
 error: the compiler unexpectedly panicked. this is a bug.
 
```
I assume that's a bug and the test suite is supposed to pass cleanly without changing any of the files?

oli-obk:
yea... that's very weird, but probably due to platform differences between your system and mine. we'll need to filter away the `.c` and `.S` files, too. I may have just filtered away backtraces for `.rs` files 🙃

[...]

PR #171, opened 2023-09-26
fix 'cargo test'

RalfJung:
Fixes https://github.com/oli-obk/ui_test/issues/168


RalfJung:
I think what might be happening is that output normalization removes the part of the diff that would show the difference...^^

RalfJung:
Ah, got it. :)

PR #180, opened 2023-10-12
Remove `$DIR` replacement of test's parent directory

Alexendoo:
I've often wanted to ditch the `$DIR` replacement to make paths clickable in the terminal, this would allow that

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1196,7 +1196,7 @@ fn check_output(
     revision: &str,
 ) -> PathBuf {
     let target = config.target.as_ref().unwrap();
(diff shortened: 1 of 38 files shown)


oli-obk:
Will that work in clippy and miri CI? If so, should we just ditch it in general?

Alexendoo:
Yeah it should:tm: work fine as it's replacing a relative path. The biggest thing is it will be a hefty diff when upgrading `ui_test`

Ditching it in general sounds good to me, it could be recreated without a new `Match` by using the per test config callback if it's desired somewhere

PR #182, opened 2023-11-06
Add option to set minimum "mandatory" message level

GuillaumeGomez:
It is needed for clippy because we only want to set "error" message annotations as mandatory and not fail in case lower level are present in the file.

The reason is that currently, if a lower level message annotation is present, the "general level" of annotation becomes the lowest present in the test file, which is not something we want.

Needed for https://github.com/rust-lang/rust-clippy/pull/11421.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,6 +1,8 @@
 use regex::bytes::Regex;
 
-use crate::{dependencies::build_dependencies, CommandBuilder, Filter, Match, Mode, RustfixMode};
+use crate::{
+    dependencies::build_dependencies, CommandBuilder, Filter, Level, Match, Mode, RustfixMode,
+};
 pub use color_eyre;
 use color_eyre::eyre::Result;
 use std::{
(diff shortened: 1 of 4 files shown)


oli-obk:
Why doesn't `//@require_annotations_for_level ERROR` satisfy your use case?

GuillaumeGomez:
It would need to be added to all clippy tests with mixed levels. It's a possibility, but I found it easier to set it directly for all tests (since it's already possible to use this option...) rather than adding it in hundred of existing ones. It will also makes things simpler for potential future new tests with mixed annotation levels if users don't need to know about this annotation (which, I precise, does exactly what we need).

Alexendoo:
While there are hundreds of test files with them in currently, almost all of them contain only `NOTE: X implied by Y` or `NOTE: X on by default` which would be better off being removed

We could bring it up at a meeting, but my impression is we will remove all the `NOTE` annotations before making it mandatory anyway. If there is a test that does want to use them it could use `//@require_annotations_for_level`, or it could just specifiy the other `NOTE`s in the file

GuillaumeGomez:
If it's only this kind of note messages, then better remove them indeed. Just would be sad to prevent adding other annotations but if it's what clippy devs want, fine by me. :)

oli-obk:
The way we should implement something like this in ui_test is to provide a default test config. That's a major refactoring that I probably have lots of opinions on, so I'd like to not just review impls but be part of the design and brain storm phase.

GuillaumeGomez:
We're not in a hurry so more discussions and brain storming are very welcome.

oli-obk:
This is resolved by https://github.com/oli-obk/ui_test/pull/183

PR #188, opened 2024-01-11
bump annotate-snippets to 0.10

klensy:
* [ ] Add tests
* [x] Add CHANGELOG.md entry

### DIFF
--- a/src/status_emitter.rs
+++ b/src/status_emitter.rs
@@ -1,9 +1,6 @@
 //! Variaous schemes for reporting messages during testing or after testing is done.
 
-use annotate_snippets::{
-    display_list::{DisplayList, FormatOptions},
-    snippet::{Annotation, AnnotationType, Slice, Snippet, SourceAnnotation},
-};
+use annotate_snippets::{Annotation, AnnotationType, Renderer, Slice, Snippet, SourceAnnotation};
 use bstr::ByteSlice;
 use colored::Colorize;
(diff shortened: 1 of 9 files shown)


src/status_emitter.rs
```diff
@@ -642,13 +639,9 @@ fn create_error(
             })
             .collect(),
         footer: vec![],
-        opt: FormatOptions {
-            color: colored::control::SHOULD_COLORIZE.should_colorize(),
```
klensy, line 646: colors here

oli-obk:
> FIXME: currently there no test fail but should, as previously colors can be switched on\off but not now.

is it detected automatically now? if so, that is fine. If it becomes an issue in tests we can always use a crate for stripping terminal colors

src/status_emitter.rs
klensy: From docs (if i understand correctly, https://docs.rs/annotate-snippets/latest/annotate_snippets/struct.Renderer.html#method.plain), `Renderer::plain()` is simple not colored output, while `Renderer::styled()` is some default color scheme.

klensy:
> is it detected automatically now? if so, that is fine. If it becomes an issue in tests we can always use a crate for stripping terminal colors

Nope, with current PR it always not colored, that why fixme added.

oli-obk:
Thanks! With your link and info I was able to get colors back

klensy:
Thanks, forgot about this PR.

PR #196, opened 2024-01-26
Fix cross-compilation

daxpedda:
When trying to cross-compile to `wasm32-unknown-unknown` (for compilation fail tests only) I realized that the builder command was not actually taking the target into account: it was still building for the host target.

Simply put, the builder command was just not taking [`Config::target`] into account.

Dependencies were already compiled to the correct target, but unfortunately I'm not familiar at all with `ui_test` so there might be something else broken that I didn't try/encounter yet.

Happy to add a test if pointed into the appropriate direction.


daxpedda:
Fixed the issues brought up in CI and added a comment to the implementation.

oli-obk:
> Happy to add a test if pointed into the appropriate direction.

I haven't created any cross compile tests yet. It'll probably be annoying to do, so don't worry about it

oli-obk:
Hmm... I think I'll need to add tests first. I worry this will interact badly with proc macro aux builds (which need to happen for the host)

oli-obk:
> Hmm... I think I'll need to add tests first. I worry this will interact badly with proc macro aux builds (which need to happen for the host)

jup it did interact badly 😆 fixed it now, but still can't get CI to actually work with cross compilation

oli-obk:
Oh 🤷 I'm just not gonna test it in CI for now

daxpedda:
Cool, thank you!

PR #198, opened 2024-02-13
Use default comments when building auxiliary files

Jarcho:
Aux builds use the wrong settings otherwise. (e.g. building for the 2015 edition)


Jarcho:
CI is failing due to "error: aborting due to 1 previous error" and "error: aborting due to previous error" differences in the output.

PR #199, opened 2024-02-13
Allow error comments above the error site.

Jarcho:
Clippy has some lints which take comments into account and where the error span encompasses multiple lines. Currently this requires putting the matcher many lines later and then counting the lines. e.g.

```rust
if foo {
  foo;
  bar;
  baz;
} else {
  foo;
  bar;
  baz;
} 
//~^^^^^^^^^ ERROR: same if else bodies
```

This patch allows putting the matcher above, avoiding the problem.

```rust
//~v ERROR: same if else bodies
if foo {
  foo;
  bar;
  baz;
} else {
  foo;
  bar;
  baz;
} 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-### Fixedw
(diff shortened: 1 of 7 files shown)


oli-obk:
neat!

PR #200, opened 2024-02-20
Allow `//~|` comments to preceed `//~v` comments.

Jarcho:
Fixes an oversight from when `//~v` comments were added. Current implementation flips the order required which wouldn't be backwards compatible. Supporting stacking both above and below would be trivial if you'd prefer.

This should probably error when it's ambiguous which arrow to follow, though it currently binds above in such a case. e.g.

```rust
//~^ error
//~| error2 
//~v error3
```

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -226,6 +226,15 @@ pub(crate) enum ErrorMatchKind {
     Code(Spanned<String>),
 }
 
+impl ErrorMatchKind {
+    fn span(&self) -> &Span {
+        match self {
+            Self::Pattern { pattern, .. } => &pattern.span,
+            Self::Code(code) => &code.span,
+        }
+    }
+}
+
 #[derive(Debug, Clone)]
 pub(crate) struct ErrorMatch {
     pub(crate) kind: ErrorMatchKind,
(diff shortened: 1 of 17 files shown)


oli-obk:
Oh god lol. That does make sense, yea.

I think we should also error on

```
//~^ error
//~v error3
```

And require a newline between for readability

Jarcho:
Do you want to allow them underneath?
```rust
//~vv error1
//~| error2
```

It's trivial either way.

Jarcho:
While we're on breaking changes, how about inline matching:

```rust
//~ error1
//~| error2

//~| error1
//~ error2

//~| error1
//~ error2
//~| error3
```

The first is currently accepted. Should it be? What about the other two?

Jarcho:
Current status:

Allowed:
```rust
//~| error1
//~v error2
```

Disallowed (not backwards compatible):
```rust
//~vv error1
//~| error2

//~ error1
//~| error2
```

Disallowed (backwards compatible):
```rust
//~| error1
//~ error2

//~^ error1
//~v error2

//~^ error1
//~| error2
//~| ...
//~v errorN
```

oli-obk:
This lgtm, and the breaking change is acceptable. It was not intended to be supported

Please also add a test for

```
//~ error1
//~| error2
```

oli-obk:
> It's probably unused

I just saw it in the rustc ui tests last week 😆 but we should just eliminate them. They look weird

PR #207, opened 2024-03-08
switch colored to owo-colors

klensy:
`owo-colors` already pulled via `color-eyre`, so switch to it.

`owo-colors` have weird `if_supports_color` api, so use `supports-color` directly. <strike>(`supports-color` can be bumped to 3.0 to drop `is-terminal`, but this will bump msrv to 1.70)</strike>

didn't checked if color scheme is equal to `colored`.

* [ ] Add tests
* [ ] Add CHANGELOG.md entry

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,10 +28,42 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.14"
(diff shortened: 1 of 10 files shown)


klensy:
Ok, slapping `anstream` allows tests to pass, but pulls a lot of deps, so left it as draft for now.

PR #209, opened 2024-03-27
Fix docs for `{ignore,only}-target-*` directive differences with compiletest

jieyouxu:
`ignore-target-*` and `only-target-*` does not, in fact, "operate solely on the triple, instead of supporting things like `macos`". It actually supports things like `macos` AFAICT because ui_test seems to match the `*` as a substring of target triples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,9 @@ their command specifies, or the test will fail without even being run.
 
 ## Significant differences to compiletest-rs
 
-* `ignore-target-*` and `only-target-*` operate solely on the triple, instead of supporting things like `macos`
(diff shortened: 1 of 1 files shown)


oli-obk:
It is phrased poorly, but correct. Instead of compiletest's support for `unix` and `macos`, we require you to specify parts of the triple, e.g. `apple`

jieyouxu:
> It is phrased poorly, but correct. Instead of compiletest's support for `unix` and `macos`, we require you to specify parts of the triple, e.g. `apple`

I fixed the wording to reflect this fact, that "special collections" like "unix"/"macos" from compiletest is not supported.

PR #235, opened 2024-05-25
Support `-Zbuild-std`

### LINKED ISSUE #197: `-Zbuild-std` support
Passing `-Zbuild-std` to `Config::dependency_builder` does the right thing, but unfortunately `Config::program` requires more help to register the right dependencies.

The call from Cargo adds these args which are missing from `ui_test`:
```
--extern 'noprelude:alloc=/home/rust/test-ui_test/target/debug/deps/liballoc-b1a1c1cd8f555a51.rlib'
--extern 'noprelude:compiler_builtins=/home/rust/test-ui_test/target/debug/deps/libcompiler_builtins-644694813e268ea1.rlib'
--extern 'noprelude:core=/home/rust/test-ui_test/target/debug/deps/libcore-5c3d328d2c56cef9.rlib'
--extern 'noprelude:panic_abort=/home/rust/test-ui_test/target/debug/deps/libpanic_abort-b22430a779820425.rlib'
--extern 'noprelude:panic_unwind=/home/rust/test-ui_test/target/debug/deps/libpanic_unwind-d0ae2f1d6268e79a.rlib'
--extern 'noprelude:proc_macro=/home/rust/test-ui_test/target/debug/deps/libproc_macro-c9c7a2404ad05f6b.rlib'
--extern 'noprelude:std=/home/rust/test-ui_test/target/debug/deps/libstd-1aa5d2ec05eb93ba.rlib'
--extern 'noprelude:test=/home/rust/test-ui_test/target/debug/deps/libtest-eed65dbc0bf35809.rlib'
```

Cargo has a bit of machinery in place to make this happen:
https://github.com/rust-lang/cargo/blob/3a72bf343b41da9d1676ad8f7b5daea2824d760e/src/cargo/core/compiler/unit_dependencies.rs#L150-L198

So basically it boils down to:
- Knowing that we did/want to build Std.
- Finding the Std root folder (installed via Rustup as the `rust-src` component).
- Calculating the dependencies (which should be the same as what `ui_test` already does).
- Add them correctly to the `rustc` call.

I'm currently looking into porting some of this code into `ui_test` and wanted to know if this in scope.

daxpedda:
As discussed in https://github.com/oli-obk/ui_test/issues/197#issuecomment-2124794280.

Fixes #197.

### DIFF
--- a/src/dependencies.rs
+++ b/src/dependencies.rs
@@ -90,6 +90,14 @@ fn build_dependencies_inner(
         build.arg(format!("--target={target}"));
     }
 
+    if let Some(packages) = &info.build_std {
+        if packages.is_empty() {
+            build.arg("-Zbuild-std");
+        } else {
+            build.arg(format!("-Zbuild-std={packages}"));
+        }
+    }
+
     // Reusable closure for setting up the environment both for artifact generation and `cargo_metadata`
     let set_locking = |cmd: &mut Command| {
         if let OutputConflictHandling::Error = config.output_conflict_handling {
@@ -166,7 +174,7 @@ fn build_dependencies_inner(
                     .target
                     .crate_types
                     .iter()
-                    .all(|ctype| !matches!(ctype.as_str(), "proc-macro" | "lib"))
+                    .all(|ctype| !matches!(ctype.as_str(), "proc-macro" | "lib" | "rlib"))
                 {
                     continue;
                 }
@@ -174,7 +182,10 @@ fn build_dependencies_inner(
                     import_paths.insert(filename.parent().unwrap().into());
                 }
                 let package_id = artifact.package_id;
-                if let Some(prev) = artifacts.insert(package_id.clone(), Ok(artifact.filenames)) {
+                if let Some(prev) = artifacts.insert(
+                    package_id.clone(),
+                    Ok((artifact.target.name, artifact.filenames)),
+                ) {
                     artifacts.insert(
                         package_id.clone(),
                         Err(format!(
@@ -252,7 +263,7 @@ fn build_dependencies_inner(
             .unwrap();
 
         // Then go over all of its dependencies
-        let dependencies = root
+        let mut dependencies = root
             .dependencies
             .iter()
             .filter(|dep| matches!(dep.kind, DependencyKind::Normal))
@@ -284,7 +295,7 @@ fn build_dependencies_inner(
                 let id = &package.id;
                 // Return the name chosen in `Cargo.toml` and the path to the corresponding artifact
                 match artifacts.remove(id) {
-                    Some(Ok(artifacts)) => Some(Ok((name.replace('-', "_"), artifacts))),
+                    Some(Ok((_, artifacts))) => Some(Ok((name.replace('-', "_"), artifacts))),
                     Some(Err(what)) => Some(Err(Errored {
                         command: Command::new(what),
                         errors: vec![],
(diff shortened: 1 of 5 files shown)


src/dependencies.rs
```diff
@@ -78,6 +78,26 @@ fn cfgs(config: &Config) -> Result<Vec<Cfg>, Errored> {
     Ok(cfgs)
 }
 
+pub(crate) fn has_build_std(info: &DependencyBuilder) -> bool {
```
oli-obk, line 81: I think it would be simpler to have a `build_std` bool field on `DependencyBuilder` and manually inject the `-Z` flags instead of expecting the user to supply them and have us detect them

oli-obk:
is there any reasonably small test that we can add here so I don't accidentally make it stop working in the future?

src/dependencies.rs
daxpedda: This couldn't be a `bool` because sometimes its relevant to pass arguments like [`-Zbuild-std=std,panic_abort`](https://doc.rust-lang.org/1.78.0/cargo/reference/unstable.html#build-std).

But I could instead make it an `Option<String>`?

I think another motivation to not add a configuration is that when https://github.com/rust-lang/wg-cargo-std-aware/issues/20 is addressed, we wouldn't need a configuration either. 

Let me know how you want me to proceed.

daxpedda:
> is there any reasonably small test that we can add here so I don't accidentally make it stop working in the future?

Would love to, but I would need some guidance to do that, would you put the test in integration or unit tests?
I see that you didn't manage adding a test in https://github.com/oli-obk/ui_test/pull/196, which I would have drawn inspiration from.

src/dependencies.rs
oli-obk: > But I could instead make it an `Option<String>`?

yea that seems good



> I think another motivation to not add a configuration is that when [rust-lang/wg-cargo-std-aware#20](https://github.com/rust-lang/wg-cargo-std-aware/issues/20) is addressed, we wouldn't need a configuration either.

I'm doing regular major bumps anyway, seems fine to do another once that lands (or just mark it as deprecated then)

oli-obk:
> I see that you didn't manage adding a test in #196, which I would have drawn inspiration from.

that one was specifically for cross compilation, which is a bit annoying. Here we can just build-std for the host itself. You can probably just add a copy of the config in https://github.com/oli-obk/ui_test/blob/9f5fec58ca5f62c21b59b35ea6b34eeb4681251c/tests/integrations/basic/tests/ui_tests.rs#L39 with build-std enabled

src/dependencies.rs
daxpedda: Done.

After trying it out though I have to say its a bit weird, because now **if** the user supplies the `-Zbuild-std` argument without enabling the `DependencyBuilder::build_std` option, it won't work.

daxpedda:
> > I see that you didn't manage adding a test in #196, which I would have drawn inspiration from.
> 
> that one was specifically for cross compilation, which is a bit annoying. Here we can just build-std for the host itself. You can probably just add a copy of the config in
> 
> https://github.com/oli-obk/ui_test/blob/9f5fec58ca5f62c21b59b35ea6b34eeb4681251c/tests/integrations/basic/tests/ui_tests.rs#L39
> with build-std enabled

I had to use `basic-bin` instead, because `-Zbuild-std` and proc-macros are a bit weird.
The remaining issues are:
- We need to add a nightly configuration to the CI and run this test only when running nightly.
- Other tests should not run in nightly because their output will be different.

I think it might be best to add a separate integration test entirely?

oli-obk:
> I think it might be best to add a separate integration test entirely?

oof yea 😆 that makes sense

daxpedda:
> > I think it might be best to add a separate integration test entirely?
> 
> oof yea 😆 that makes sense

So I adjusted the integration test to take an env variable to set the folder and then just copied `basic-bin` with the appropriate `build_std` addition.

I think I could instead make a simple test using `ui_test` without the whole machinery in place to check the Cargo output?
Let me know what you think.

oli-obk:
> I think I could instead make a simple test using `ui_test` without the whole machinery in place to check the Cargo output?
> Let me know what you think.

right, we only care that it works, not what the output is

daxpedda:
> > I think I could instead make a simple test using `ui_test` without the whole machinery in place to check the Cargo output?
> > Let me know what you think.
> 
> right, we only care that it works, not what the output is

Done, let me know if you want any other adjustments.

(sorry for the delay)

oli-obk:
Thanks! This is great!

Issue #229, opened 2024-05-14
Default output is excessively verbose


xldenis:
I've started porting my tools over to `ui_test`, but I find the default output to be excessively verbose: it systematically dumps the diff _and_ the full `stdout` / `stderr`, could we include a version of the `Text` emitter which does not dump the full file contents? My `stdout` files are often ~20-30kloc long but with < 100 lines of diff, which makes the default status emitter fairly useless.

PR #236, opened 2024-05-28
Add diff-only output on Text

xldenis:
This solves #229, I generalized the `progress` boolean to instead store the desired verbosity and added a method to construct a `diff` status emitter which just returns an empty drop guard rather than the full text. 

The todo list below is accurate: I still need to add tests.

# TODO 
* [ ] Add tests
* [ ] Add CHANGELOG.md entry


oli-obk (dismissed):

xldenis:
Ok I've added a test showing that I actually suppress the full output.

PR #241, opened 2024-06-21
drop unused tempfile

klensy:
Drop unused `tempfile`. Was added in #23, but unused (except tests).


oli-obk:
CI failure is just a tumour on dowloading clippy

PR #244, opened 2024-06-21
Create multiple `.fixed` files for diagnostics with multiple suggestions

Alexendoo:
Some clippy tests disable rustfix because they produce multiple alternative suggestions e.g. [`join_absolute_paths`](https://github.com/rust-lang/rust-clippy/blob/29cc5c691c574b54b2fcfed51c22f8bb51110919/tests/ui/join_absolute_paths.stderr). This PR allows there to be a `.fixed` file per alternative suggestion, in that example `join_absolute_paths.1.fixed` and `join_absolute_paths.2.fixed`

There are some lints that produce multiple suggestions that are intended to be combined e.g.

```
error: this binding can be a slice pattern to avoid indexing
  --> tests/ui/index_refutable_slice/if_let_slice_binding.rs:21:17
   |
LL |     if let Some(slice) = slice {
   |                 ^^^^^
   |
help: try using a slice pattern here
   |
LL |     if let Some([slice_0, ..]) = slice {
   |                 ~~~~~~~~~~~~~
help: and replace the index expressions here
   |
LL |         println!("{}", slice_0);
   |                        ~~~~~~~
```

These would have to use `//@no-rustfix` for now, but I think moving them to multipart suggestions is correct in the long run

### DIFF
--- a/src/custom_flags/rustfix.rs
+++ b/src/custom_flags/rustfix.rs
@@ -2,10 +2,11 @@
 
 use std::{
     collections::HashSet,
-    path::Path,
+    path::{Path, PathBuf},
     process::{Command, Output},
 };
 
+use rustfix::{CodeFix, Filter, Suggestion};
 use spanned::{Span, Spanned};
 
 use crate::{
@@ -13,8 +14,7 @@ use crate::{
     display,
     parser::OptWithLine,
     per_test_config::{Comments, Revisioned, TestConfig},
-    test_result::TestRun,
-    Error, Errored, TestOk,
+    Error, Errored, TestOk, TestRun,
 };
 
 use super::Flag;
@@ -49,7 +49,7 @@ impl Flag for RustfixMode {
         _cmd: &mut Command,
         output: &Output,
         build_manager: &BuildManager<'_>,
-    ) -> Result<Option<TestRun>, Errored> {
+    ) -> Result<Vec<TestRun>, Errored> {
         let global_rustfix = match config.exit_status()? {
             Some(Spanned {
                 content: 101 | 0, ..
@@ -58,7 +58,7 @@ impl Flag for RustfixMode {
         };
         let output = output.clone();
         let no_run_rustfix = config.find_one_custom("no-rustfix")?;
-        let fixed_code = if no_run_rustfix.is_none() && global_rustfix.enabled() {
+        let fixes = if no_run_rustfix.is_none() && global_rustfix.enabled() {
             fix(&output.stderr, config.status.path(), global_rustfix).map_err(|err| Errored {
                 command: format!("rustfix {}", display(config.status.path())),
                 errors: vec![Error::Rustfix(err)],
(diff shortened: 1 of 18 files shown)


src/custom_flags/rustfix.rs
```diff
@@ -53,115 +63,51 @@ impl Flag for RustfixMode {
...
+                            &format!("{}.fixed", i + 1),
+                        )
+                    })
+                    .collect()
+            }
         };
+
+        if fixed.len() != 1 {
+            // Remove an unused .fixed file
+            config.check_output(&[], &mut errors, "fixed");
```
oli-obk, line 108: this will fail to remove `.1.fixed` files. Maybe we need to use the `glob` crate to look for specific patterns of files to remove now

src/custom_flags/rustfix.rs
```diff
@@ -171,18 +117,124 @@ impl Flag for RustfixMode {
...
+
+        if fixed.is_empty() {
+            Ok(Some(cmd))
+        } else {
+            Ok(None)
         }
+    }
+}
+
+fn fix(stderr: &[u8], path: &Path, filter: rustfix::Filter) -> anyhow::Result<Vec<String>> {
```
oli-obk, line 130: For the future, when doing large refactorings like this, please split out the code movements into separate commits

tests/integrations/basic-fail/tests/ui_tests_bless.rs
```diff
@@ -46,18 +46,13 @@ fn main() -> ui_test::color_eyre::Result<()> {
...
             default_file_filter,
             default_per_file_config,
             // Avoid github actions, as these would end up showing up in `Cargo.stderr`
             status_emitter::Text::verbose(),
         );
-        match (&result, rustfix) {
-            (Ok(_), RustfixMode::Everything) => {}
-            (Err(_), RustfixMode::MachineApplicable) => {}
-            _ => panic!("invalid mode/result combo: {rustfix:?}: {result:?}"),
-        }
```
oli-obk, line 60: what's up with this?

Alexendoo: I added a test to `actual_tests_bless_yolo` that fails in the rustfix step

src/custom_flags/rustfix.rs
Alexendoo: Yeah I wasn't sure if I should do a `read_dir` or something along the lines to catch the other ones, it would be quadratic if we're checking it for every `.fixed` file

Perhaps we could check `.1.fixed`, `.2.fixed` etc until we stop finding them

src/custom_flags/rustfix.rs
Alexendoo: Sorry yeah, should've gone back to do that. Should I split it out still or leave it as is?

oli-obk: Nah it's ok, but the Rebase will be harder

src/custom_flags/rustfix.rs
oli-obk: Yea... let's leave it as is. Maybe i'll record all files in the future and check them once against the actual file list

src/custom_flags/rustfix.rs
Alexendoo: Yeah there were indeed a lot of conflicts so I ended up reapplying it manually, stuck it in its own commit while I was there

src/per_test_config.rs
```diff
@@ -416,16 +407,21 @@ impl TestConfig<'_> {
...
                 for flag in &custom.content {
-                    if let Some(result) =
-                        flag.post_test_action(self, &mut cmd, &output, build_manager)?
-                    {
-                        runs.push(result);
-                        return Ok(TestOk::Ok);
-                    }
+                    flag.post_test_action(&silent, &mut cmd, &output, build_manager)?;
                 }
             }
```
Alexendoo, line 426: I ended up making `Flag` things silent

Personally I prefer not having the `.fixed` etc files be their own line in the log, but also timing wise the spinners felt redundant as the main test's/revision's spinner and all of its sub tests would be marked as complete at once after `run_test` returned

oli-obk: Let's do that in a separate PR, as I'd rather do some experimentation, and not block this PR on said experimentation

Alexendoo: Should I have it return a Vec in this PR in that case?

oli-obk: Hmm good point. Yea a vec seems best

src/custom_flags/rustfix.rs
```diff
@@ -58,94 +58,126 @@ impl Flag for RustfixMode {
...
-            "fixed",
-        );
+        let mut errors = Vec::new();
+        let fixed_paths = match fixes.as_slice() {
+            [] => Vec::new(),
+            [single] => {
+                vec![config.check_output(single.as_bytes(), &mut errors, "fixed")]
+            }
+            _ => {
+                config.check_output(&[], &mut errors, "fixed");
```
oli-obk, line 79: this line is redundant with the `Remove an unused .fixed file` line below

PR #247, opened 2024-07-05
Silence subtests in `Flag`s

Alexendoo:
Two main reasons:

- Having log lines/spinners for each `.fixed` file feels like overkill
- The spinners all resolve at the same time as the main test making them redundant (fixable)


oli-obk:
I fixed this by nesting spinners below the main spinner and making them finish independently of each other.

PR #255, opened 2024-08-03
We handle carriage return within ui_test

oli-obk:
* [x] Add tests
* [ ] Add CHANGELOG.md entry


oli-obk:
turns out it's not needed to handle jj after all

PR #257, opened 2024-08-05
Add full readme to crate documentation

GuillaumeGomez:
Currently we need to go to the repository to have access to all the syntax information. Instead, this PR includes it directly into the crate documentation (but only when rustdoc is building).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,8 +5,7 @@
     rustc::internal
 )]
 #![deny(missing_docs)]
-
-//! A crate to run the Rust compiler (or other binaries) and test their command line output.
+#![doc = include_str!("../README.md")]
 
 use build_manager::BuildManager;
 pub use color_eyre;


oli-obk:
Uhm... needs rustfmt and possibly a dummy rustdoc comment. Looks like a bug in the missing doc lint?

src/lib.rs
```diff
@@ -5,8 +5,11 @@
     rustc::internal
 )]
 #![deny(missing_docs)]
-
-//! A crate to run the Rust compiler (or other binaries) and test their command line output.
+#![cfg_attr(doc, doc = include_str!("../README.md"))]
+#![cfg_attr(
+    not(doc),
```
oli-obk, line 10: Hmm.. is this the intended usage? Why not just unconditionally import? Or make `doc` attributes disappear without cfg(doc) in general?

GuillaumeGomez: I think it does at later stages. But if it made them disappear right after the lint pass, rustdoc wouldn't have access to them.

I didn't unconditionally import to prevent adding a watcher over the readme file because I know some people are sensible to these kind of things but if it doesn't matter for you, I can do it. It'll simplify the code too. :)

PR #271, opened 2024-09-08
replace lazy_static with std's OnceLock

klensy:
[OnceLock](https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html) stable from 1.70, which is current msrv, so use it instead (`lazy_static` still in tree),
* [ ] Add tests
* [x] Add CHANGELOG.md entry


oli-obk:
Cool thx!

PR #286, opened 2024-10-13
Export parser types

DaniPopes:
`CommentParser` is unnameable from outside this crate, meaning a user cannot declare a parser function with `fn`, only inline with a closure. I've fixed this by exporting `parser::*`, and enabled the `unreachable_pub` and `unnameable_types` to catch this and other unreachable types.


oli-obk:
Thanks! That lint is so helpful

PR #287, opened 2024-10-13
Cache an error parser regex

DaniPopes:
A regex in `rustc_stderr` gets created on every JSON stderr line, which shows up in profiles when running thousands of tests.  
Cache it with `OnceLock` (MSRV 1.70, also used elsewhere in the codebase).


oli-obk:
Oops, good catch

PR #289, opened 2024-10-14
Export default diagnostics extractors

DaniPopes:
The config `diagnostic_extractor`s provided by the `rustc` feature are not exported but reachable through `Config::{cargo,rustc}().diagnostic_extractor`; making these public would let users import the function directly instead of having it to access it through `Config` methods.
I've renamed them and moved them to the `diagnostics` module, as well as added the default one as a separate function.

My use case for this is I am testing another compiler that can emit JSON diagnostics using the rustc format, but everything else about the configuration will be different, so it doesn't really make sense to call `Config::rustc` if it wasn't for this method.

BTW I was able to migrate to this library from a compiletest-rs fork pretty easily, thanks a lot for developing this!

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -1,10 +1,17 @@
 //! Data structures for handling diagnostic output from tests.
 
-#[cfg(feature = "rustc")]
-use cargo_metadata::diagnostic::DiagnosticLevel;
+use std::path::Path;
 
-#[derive(Copy, Clone, Debug, PartialOrd, Ord, PartialEq, Eq)]
+#[cfg(feature = "rustc")]
+pub mod rustc;
+
+/// Default diagnostics extractor that does nothing.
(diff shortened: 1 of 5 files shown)


src/diagnostics.rs
```diff
@@ -2,9 +2,20 @@
 
 #[cfg(feature = "rustc")]
 use cargo_metadata::diagnostic::DiagnosticLevel;
+use std::path::Path;
+
+#[cfg(feature = "rustc")]
+mod rustc_stderr;
+#[cfg(feature = "rustc")]
+pub use rustc_stderr::{cargo_diagnostics_extractor, rustc_diagnostics_extractor};
```
oli-obk, line 10: I think we can simplify this a bit further by 
* renaming the module to `rustc`
* making it public instead of reexporting things
* moving the cfged impl into it, too

oli-obk (commented):
Good idea. It should be possible to build a rustc config manually without the rustc constructor function. There's still more I need to expose, but this is an important piece

Issue #290, opened 2024-10-30
Add support for revisions without rustc


DaniPopes:
Currently specifying the directive `@revisions <revision>...` implicitly appends `--cfg=<revision> -Cextra-filename=<revision>` to the compile flags, assuming the command is a rustc driver.

It would be nice to be able to use this without a rustc driver for running a test multiple times with different configurations and different output files.

I was thinking to add a simple config flag that turns the implicit compile flags off, since AFAICT this is the only implicit behavior of the directive. If this is acceptable I can open a PR to implement this.

oli-obk:
Hmm... revisions are pretty special, but maybe we can still make them a `Flag`.

If we can't, we make the "command line flag adding" a `Flag` that you can either not set at all, or create your own version of if you want it to set env vars or different flags instead. The `apply` function can query the revision from the status emitter

PR #291, opened 2024-10-30
Move revision-specific args to a custom flag

DaniPopes:
Closes https://github.com/oli-obk/ui_test/issues/290 by implementing the second suggestion in https://github.com/oli-obk/ui_test/issues/290#issuecomment-2447877332.

* [x] Add tests
* [x] Add CHANGELOG.md entry

Should already be tested by existing revisions tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


DaniPopes:
Not sure if I need to bless tests because doing so makes them pass, but they also contain "DO NOT BLESS. These are meant to fail"

oli-obk:
hmm rerunning CI apparently doesn't rebase. Oh well, `main` branch CI will catch any issues (I don't forsee any)

PR #304, opened 2025-02-12
Add example for revisions

GuillaumeGomez:
I'm working on enforcing annotation in clippy ui tests and understanding revisions took me more time than I think it should have. I think adding a code example will be helpful for potential users or future me.


oli-obk:
Thanks! I guess i'll need to add more tests for using slightly wrong syntax and see what I can do to improve diagnostics

GuillaumeGomez:
Examples are always a good idea. :+1:

Issue #144, opened 2023-08-19
Some content missing from a `.stderr` diff output


Alexendoo:
On the current clippy (https://github.com/rust-lang/rust-clippy/commit/7408f1d231a5e223a2ab01032ffd1778bcb6e0f1) run

```bash
git apply <<EOF
diff --git i/tests/ui/unwrap_or_else_default.rs w/tests/ui/unwrap_or_else_default.rs
index afacedf17..be44b7d7d 100644
--- i/tests/ui/unwrap_or_else_default.rs
+++ w/tests/ui/unwrap_or_else_default.rs
@@ -130,4 +130,17 @@ fn method_call_with_deref() {
     let _ = inner_map.entry(0).or_insert_with(Default::default);
 }
 
+fn no_suggested_method() {
+    struct Entry<T>(T);
+
+    impl<T> Entry<T> {
+        fn or_insert_with(&mut self, default: impl FnOnce() -> T) -> &mut T {
+            &mut self.0
+        }
+    }
+
+    let mut entry = Entry(1);
+    entry.or_insert_with(Default::default);
+}
+
 fn main() {}
EOF
```

`TESTNAME=unwrap_or_else_default cargo uitest` gives the diff

```diff
actual output differed from expected
Execute `cargo uibless` to update `tests/ui/unwrap_or_else_default.stderr` to the actual output
--- tests/ui/unwrap_or_else_default.stderr
+++ <stderr output>
 error: use of `unwrap_or_else` to construct default value
   --> $DIR/unwrap_or_else_default.rs:46:14
... 94 lines skipped ...
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `or_default()`
 
-error: aborting due to 16 previous errors
+error: use of `or_insert_with` to construct default value
 
+error: aborting due to 17 previous errors
+
```

The diagnostic is partly missing from the diff, it's present in the `full stderr:` section and blessing the test and running `git diff` shows the expected diff:

```diff
diff --git i/tests/ui/unwrap_or_else_default.stderr w/tests/ui/unwrap_or_else_default.stderr
index 40023cf3b..a36dd29a3 100644
--- i/tests/ui/unwrap_or_else_default.stderr
+++ w/tests/ui/unwrap_or_else_default.stderr
@@ -96,5 +96,11 @@ error: use of `or_insert_with` to construct default value
 LL |     let _ = inner_map.entry(0).or_insert_with(Default::default);
[...]

PR #308, opened 2025-03-03
Fix missing lines in diff output

Alexendoo:
Fixes https://github.com/oli-obk/ui_test/issues/144

`Replaced` can contain a different number of lines in `l` and `r`, so fall back to a plain diff output in that case


oli-obk:
Aaaaa thanks!!

PR #310, opened 2025-03-11
Add `ICE` level parsing

GuillaumeGomez:
Needed to finish to make ui test annotations mandatory in clippy.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -36,7 +36,7 @@ impl std::str::FromStr for Level {
             "HELP" | "help" => Ok(Self::Help),
             "NOTE" | "note" => Ok(Self::Note),
             "failure-note" => Ok(Self::FailureNote),
-            "error: internal compiler error" => Ok(Self::Ice),
+            "ICE" | "ice" => Ok(Self::Ice),
             _ => Err(format!("unknown level `{s}`")),
         }
     }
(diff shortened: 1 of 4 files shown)


src/diagnostics.rs
```diff
@@ -36,7 +36,7 @@ impl std::str::FromStr for Level {
             "HELP" | "help" => Ok(Self::Help),
             "NOTE" | "note" => Ok(Self::Note),
             "failure-note" => Ok(Self::FailureNote),
-            "error: internal compiler error" => Ok(Self::Ice),
+            "ICE" | "ice" | "error: internal compiler error" => Ok(Self::Ice),
```
GuillaumeGomez, line 39: I'm not sure if it's supposed to be like this but we can never have `"error: internal compiler error"` here as far as I know since the content after the `:` is removed when parsing the pattern. So instead, I think it makes sense to add a new `ICE` level to handle this case.

oli-obk: If there are no affected tests, yes.

Also add a test for the new annotation. There should be an existing Ztreat-err-as-bug test. If not, add one

GuillaumeGomez: I was sure there was a `//@ failure-status: 101` command but apparently not. Might be worth as a follow-up. Anyway, added the test and removed the extra text matching.

GuillaumeGomez:
And fixed CI. :)

src/diagnostics.rs
oli-obk: > was sure there was a `//@ failure-status: 101` command bu

Yea I should probably add that

oli-obk:
Thanks! Wanted this forever

GuillaumeGomez:
Btw, can you make a release please? I need it to finish my clippy PR.

oli-obk:
Ah, will do later today. If you want this in the future, bump the version in your PR, too. Then I can release from my phone

PR #312, opened 2025-04-13
Change default fixed suffix to fixed.rs

### LINKED ISSUE #288: Change `.fixed` files to `.fixed.rs`
While implementing UI tests for a custom linter, [a reviewer noted that it would easier to review `rustfix`'d files if they had a `.rs` extension](https://github.com/TheBevyFlock/bevy_cli/pull/128#discussion_r1798345558), since that would enable code highlighting when reviewing on Github or other code editors. This would be useful, but would be a breaking change for existing users.

I believe the extension for `rustfix`'d files is currently hardcoded here:

https://github.com/oli-obk/ui_test/blob/5b7b0b2bd159814ba1cd8d1c4dfa0830ea38755e/src/custom_flags/rustfix.rs#L73-L79

Additionally, this would probably require [`default_file_filter()`](https://github.com/oli-obk/ui_test/blob/5b7b0b2bd159814ba1cd8d1c4dfa0830ea38755e/src/lib.rs#L104) to be updated to skip `.fixed.rs` files.

Thanks!

michaelciraci:
Closes #288 

This changes the default fixed suffix from `fixed` to `fixed.rs` for IDE rendering.

* [x] Add tests
* [x] Add CHANGELOG.md entry


michaelciraci:
What if we let the fixed suffix be either `.fixed` or `.fixed.rs`? Then it won't break anything that's currently working, but maintainers can change the suffix when it works for them to support the IDE rendering. If a maintainer doesn't care about IDE rendering, they don't ever have to bless `.fixed`

Issue #315, opened 2025-04-22
Add libtest JSON emitter


kirtchev-adacore:
As discussed in the [#t-infra/bootstrap > x test src/tools/clippy and JSON output](https://rust-lang.zulipchat.com/#narrow/channel/326414-t-infra.2Fbootstrap/topic/x.20test.20src.2Ftools.2Fclippy.20and.20JSON.20output) Zulip thread, oli suggested:

> Adding json support for ui_test shouldn't be too hard (famous last words). I wrote the entire system emitter-agnostic. So you "just" need to implement the various emitter traits for a new json emitter struct. If you want to upstream it to ui_ test I'll happily maintain it, so there's only a one time cost for you (implementing, adding tests and likely addressing a round or two of review feedback)

For the record, I looked into creating a `ui_test`-to-`libtest` adapter, where implementations of the various emitter traits would convert the `ui_test` types to `libtest` types and just use the `libtest` JSON emission circuitry, however `libtest` does not expose most of the relevant traits and types in its public API.

oli-obk:
yea we'll need to add the types ourselves to ui_test. Unless `cargo_metadata` has them already, I'm not sure how much stuff we added to cargo_metadata

Zalathar:
For comparison: When implementing libtest-compatible JSON output in compiletest (https://github.com/rust-lang/rust/pull/139660), I ended up writing the JSON emitters by hand, because I found the equivalent serde declarations to be much harder to write and read.

oli-obk:
Yea, I don't think a serde style serialization is possible here in general the way the emitter traits work. The outer layers will need to be custom, or you need to store everything in one big data structure and serialize at the end

kirtchev-adacore:
I have been looking at the `status_emitter` traits and the implementation in `text` since it will be the closest implementation to the libtest JSON, and I am scratching my head a bit.

1. Please check my understanding of how `ui_test` works with respect to emitters:
    1. `ui_test` calls `StatusEmitter::register_test()` for each discovered test, getting back a `TestStatus`.
    2. The mapping between a test and its `TestStatus` is maintained by `ui_test`.
    3. Once the test completes, `ui_test` calls `TestStatus::done()` for the corresponding `TestStatus`.
    4. When is `TestStatus::failed_test()` called with respect to `TestStatus::done()`?
    5. What is this "drop guard" mentioned in the comment of `TestStatus::failed_test()`? What does it do?
[...]

PR #316, opened 2025-04-24
Add libtest JSON emitter

kirtchev-adacore:
This PR adds a status emitter which outputs the libtest JSON format, as specified in `library/test/src/formatters/json.rs`.

fixes https://github.com/oli-obk/ui_test/issues/315

* [x] Add tests
* [x] Add CHANGELOG.md entry


src/status_emitter.rs
```diff
@@ -6,14 +6,34 @@ use std::{
...
 #[cfg(feature = "gha")]
 pub use gha::*;
 #[cfg(feature = "gha")]
 mod gha;
+pub use libtest_json::*;
+mod libtest_json;
+pub use text::*;
+mod text;
+
+// The testing framework employs the implementations of the various emitter traits
```
oli-obk, line 21: turn this into module level docs

src/status_emitter.rs
```diff
@@ -228,6 +248,24 @@ impl<T: StatusEmitter> StatusEmitter for (T,) {
     }
 }
 
+impl<T: StatusEmitter + Send> StatusEmitter for (Arc<T>,) {
```
oli-obk, line 251:
```suggestion
impl<T: StatusEmitter + Send> StatusEmitter for Arc<T> {
```

should be the right thing, then the tuple impl will pick it up automatically

src/status_emitter/text.rs
```diff
@@ -343,9 +343,12 @@ impl Text {
 
 impl From<Format> for Text {
     fn from(format: Format) -> Self {
+        // Not sure what to do here. The architecture allows for multiple emitter
+        // implementations, but the code assumes that there is only implementation,
+        // namely Text.
```
oli-obk, line 348: yea this impl should be removed. Probably just in favor of an `From<Format> for Arc<dyn StatusEmitter>` impl or just some free function

src/config/args.rs
```diff
@@ -76,8 +78,9 @@ impl Args {
                 // We ignore this flag for now.
             } else if let Some(format) = parse_value("--format", &arg, &mut iter)? {
                 self.format = match &*format {
-                    "terse" => Format::Terse,
+                    "libtest-json" => Format::LibtestJSON,
```
oli-obk, line 81: hmm why not just `json`?

src/status_emitter/libtest_json.rs
```diff
@@ -0,0 +1,227 @@
...
+        filtered: usize,
+        aborted: bool,
+    ) -> Box<dyn Summary> {
+        let status = if aborted || failed > 0 {
+            String::from("failed")
+        } else {
+            String::from("ok")
+        };
+
+        emit_suite_start(failed + ignored + succeeded);
```
oli-obk, line 133: 😆 shouldn't this be emitted first like other tests do it? Which is like the whole problem with counting tests

oli-obk: Either do not emit a count, or add logic to ui_test to do a dry run that just collects the number of tests

src/status_emitter.rs
kirtchev-adacore: My colleague actually tried this, and it did not work (I don't have the exact errors at hand). Instead he got this to work (see `patch.txt`).
[patch.txt](https://github.com/user-attachments/files/19889486/patch.txt)

src/config/args.rs
kirtchev-adacore: This was intentional, in case you would like to add "full" JSON support at some point.

Note that some of the events generated by the testing framework are not emitted, hence I was apprehensive about calling this "JSON". Up to you really.

src/status_emitter/libtest_json.rs
kirtchev-adacore: It turns out that as long as the event is emitted at some point, this will satisfy our use case.

If we want to be chronologically correct, then indeed, a "suite start" has to be the first event emitted. I am not sure whether adding dry run logic is worth the complexity and maintenance burden. (The number of tests is important!)

src/status_emitter/libtest_json.rs
```diff
@@ -0,0 +1,227 @@
...
+    let triaged_name = EscapedString(name);
+
+    // Adapted from test::formatters::json::write_test_start().
+    println!(r#"{{ "type": "test", "event": "started", "name": "{triaged_name}" }}\n"#);
+}
+
+// Adapted from test::formatters::json.
+/// A formatting utility used to print strings with characters in need of escaping.
+/// Base code taken form `libserialize::json::escape_str`.
+struct EscapedString<S: AsRef<str>>(S);
```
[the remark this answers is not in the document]
kirtchev-adacore: The use of manually created JSON strings was intentional, as I think it will minimize the maintenance burden.

I think it will be easier to just look at the corresponding `libtest` functions ([write_run_start](https://github.com/rust-lang/rust/blob/master/library/test/src/formatters/json.rs#L97) for example), and determine whether something has changed. Zalatar also favored this approach (he [mentions](https://github.com/oli-obk/ui_test/issues/315#issuecomment-2821120463) this under the issue).

Alternatively, we are looking at a type hierarchy similar to [this one](https://github.com/rust-lang/rust/blob/master/src/bootstrap/src/utils/render_tests.rs#L433) in `bootstrap/src/utils/render_tests.rs`. The types would need to be a subset of those.

Up to you.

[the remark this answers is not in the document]
oli-obk: Yea I guess we can keep it this way. I'd still like to reuse the existing json string escaping, even if a bit roundabout to use. basically run `serde_json::to_string(&the_string)` on the strings instead of reimplementing json escape logic here

src/status_emitter/libtest_json.rs
oli-obk: > (The number of tests is important!)

I really question this factoid 😆 All you are doing is sanity checking the test suite on whether it correctly precomputed the number of tests. You even completely subverted this sanity check here, as you are not emitting the number of tests after running the tests. So I would recommend to just stop relying on this number for the suite start and *only* checking the number at the suite end (and checking that against the individual test jsons you got)

src/config/args.rs
oli-obk: I have few qualms changing the json output to only mostly match libtest-json and to have more data inside that libtest doesn't emit ;) Let's just use `json`

src/status_emitter.rs
oli-obk: you can paste diffs rendered inline:

```diff
diff --git a/src/status_emitter.rs b/src/status_emitter.rs
index f3e8962..0cef7ec 100644
--- a/src/status_emitter.rs
+++ b/src/status_emitter.rs
@@ -6,7 +6,7 @@ use std::{
     fmt::Debug,
     panic::RefUnwindSafe,
     path::{Path, PathBuf},
-    sync::{Arc, Mutex},
+    sync::Arc,
 };
 pub mod debug;
 #[cfg(feature = "gha")]
@@ -36,7 +36,7 @@ mod text;
 // Summary::test_failure().
 
 /// A generic way to handle the output of this crate.
-pub trait StatusEmitter: Sync + RefUnwindSafe {
+pub trait StatusEmitter: Send + Sync + RefUnwindSafe {
     /// Invoked the moment we know a test will later be run.
     /// Useful for progress bars and such.
     fn register_test(&self, path: PathBuf) -> Box<dyn TestStatus + 'static>;
@@ -203,7 +203,7 @@ impl<T: TestStatus, U: TestStatus> TestStatus for (T, U) {
     }
 }
 
-impl<T: StatusEmitter + Send, U: StatusEmitter> StatusEmitter for (Arc<T>, U) {
+impl<T: StatusEmitter, U: StatusEmitter> StatusEmitter for (T, U) {
     fn register_test(&self, path: PathBuf) -> Box<dyn TestStatus> {
         Box::new((
             self.0.register_test(path.clone()),
@@ -248,9 +248,9 @@ impl<T: StatusEmitter> StatusEmitter for (T,) {
     }
 }
 
-impl<T: StatusEmitter + Send> StatusEmitter for (Arc<T>,) {
+impl StatusEmitter for Arc<dyn StatusEmitter> {
     fn register_test(&self, path: PathBuf) -> Box<dyn TestStatus> {
-        self.0.register_test(path.clone())
+        (**self).register_test(path.clone())
     }
 
     fn finalize(
@@ -261,8 +261,7 @@ impl<T: StatusEmitter + Send> StatusEmitter for (Arc<T>,) {
         filtered: usize,
         aborted: bool,
     ) -> Box<dyn Summary> {
-        self.0
-            .finalize(failures, succeeded, ignored, filtered, aborted)
+        (**self).finalize(failures, succeeded, ignored, filtered, aborted)
     }
 }

```

src/status_emitter/libtest_json.rs
oli-obk: This will break the test suite renderer in https://github.com/rust-lang/rust/blob/8f274491a172e635abc447eaa44438e9239b0f05/src/bootstrap/src/utils/render_tests.rs#L310 and erase all recorded tests before printing that none have been executed

Out of different reasons I started messing with the compiletest suite in rustc, and it turns out it does a lot of work ahead of time just to print a reliable number (like enough work to just run 50 tests completely from start to finish in the same time). And I found a bunch of comments hinting at things being this way just for legacy reasons. So I started a discussion in https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/compiletest.20tuning/with/515040266, which may result in us changing what compiletest generally outputs

src/status_emitter.rs
oli-obk: > My colleague actually tried this, and it did not work

I copy pasted the `Box<T>` impl to an `Arc<T>` impl and everything still compiled. 

```rust
impl<T: StatusEmitter + ?Sized + Send> StatusEmitter for Arc<T> {
    fn register_test(&self, path: PathBuf) -> Box<dyn TestStatus> {
        (**self).register_test(path)
    }

    fn finalize(
        &self,
        failures: usize,
        succeeded: usize,
        ignored: usize,
        filtered: usize,
        aborted: bool,
    ) -> Box<dyn Summary> {
        (**self).finalize(failures, succeeded, ignored, filtered, aborted)
    }
}
```

Why do you need an `Arc` anyway? clippy and miri were using `Box` just fine so far.

src/status_emitter.rs
kirtchev-adacore: I am back from vacation, and once again looking at the PR.

> turn this into module level docs

Will do!

src/status_emitter/text.rs
kirtchev-adacore: OK, I will remove it.

src/config/args.rs
kirtchev-adacore: OK, `json` it is then!

src/status_emitter/libtest_json.rs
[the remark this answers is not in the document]
kirtchev-adacore: OK, I will use `serde_json::to_string(&the_string)`.

src/status_emitter.rs
kirtchev-adacore: Note that up until now, object `text` in
```
    let text = match args.format {
        Format::Terse => status_emitter::Text::quiet(),
        Format::Pretty => status_emitter::Text::verbose(),
    };
```
had type `Text`. Now that we have a new `StatusEmitter`, we need a `dyn StatusEmitter` object. When this object is wrapped in a `Box`, I get the following error:
```
error[E0277]: `dyn status_emitter::StatusEmitter` cannot be sent between threads safely
   --> src/lib.rs:103:9
    |
99  |       run_tests_generic(
    |       ----------------- required by a bound introduced by this call
...
103 | /         (
104 | |             emitter,
105 | |             #[cfg(feature = "gha")]
106 | |             status_emitter::Gha::<true> { name },
107 | |         ),
    | |_________^ the trait `Send` is not implemented for `Unique<dyn status_emitter::StatusEmitter>`
```
This led me to believe (and I am probably wrong) that using `Arc` instead of `Box` will resolve the compilation error.

kirtchev-adacore: After much fiddling, I think the fix is to add `Send` to the `pub trait StatusEmitter`. Then I can directly use a `Box`.

src/status_emitter.rs
kirtchev-adacore: Done.

src/status_emitter.rs
kirtchev-adacore: Done.

src/status_emitter/text.rs
kirtchev-adacore: Done.

src/config/args.rs
kirtchev-adacore: Done.

src/status_emitter/libtest_json.rs
kirtchev-adacore: Done.

src/status_emitter/libtest_json.rs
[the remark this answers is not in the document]
kirtchev-adacore: Done.

kirtchev-adacore:
Oli, could you advise on how to write tests for this PR? I was looking into `tests`, and here is what I understood (which is most likely wrong or incomplete):

`tests/integration.rs` is the only place which mentions output formatting, so I assume that this is the testing entry point. The `Args` returned by `Args::test()` are parsed from the command line, so I assume I would need to pass `--format=json`, but where?

`tests/integration.rs` then calls `run_tests_generic()` on what I assume is the `tests/integrations` directory as a testing root.

And from here on, I have no idea what is going on. I don't seem to find any expected output or oracles...

kirtchev-adacore:
I struggled with testing and ended up copying and adapting `integrations/basic` and `integrations/basic-fail` to work with the JSON emitter. I noticed a couple of issues:

1. The testing framework appears to be either running or reporting on (the more likely) a single test several times:
```
{ "type": "test", "event": "ok", "name": "tests/actual_tests/bad_pattern.rs" }
{ "type": "test", "event": "failed", "name": "tests/actual_tests/bad_pattern.rs", . . .}
. . .
{ "type": "test", "event": "ok", "name": "tests/actual_tests/executable.rs" }
{ "type": "test", "event": "ok", "name": "tests/actual_tests/executable.rs" }
```
2. `serde_json::to_string()` does not properly escape `{` and `}`. If the `stderr` or `stdout` contain these characters, the JSON ends up being malformed. I will have to check what `EscapedString` does with these characters.

src/status_emitter/json.rs
```diff
@@ -0,0 +1,157 @@
...
+                Ok(TestOk::Ignored) => String::from("ignored"),
+                Ok(TestOk::Ok) => String::from("ok"),
+                Err(_) => String::from("failed"),
+            }
+        };
+        let error_output = if let Err(errored) = result {
+            let command = &errored.command;
+            let stderr = errored.stderr.to_str_lossy();
+            let stdout = errored.stdout.to_str_lossy();
+            format!(r#"---- command\n\n{command}\n\n\---- stdout\n\n\{stdout}\n\n---- stderr\n\n\{stderr}"#)
```
oli-obk, line 109: huh, test failures are not reported in json format?

src/status_emitter/json.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// The path of the test file.
+    fn path(&self) -> &Path {
+        &self.path
+    }
+
+    /// The revision, usually an empty string.
+    fn revision(&self) -> &str {
+        &self.revision
+    }
+}
```
oli-obk, line 157: missing newline at end of file

tests/integration.rs
```diff
@@ -97,7 +97,11 @@ fn main() -> Result<()> {
     config.stdout_filter(r#"exit status: 0xc0000409"#, "signal: 6 (SIGABRT)");
     config.filter("\"--target=[^\"]+\"", "");
 
-    let text = ui_test::status_emitter::Text::from(args.format);
+    let emitter: Box<dyn status_emitter::StatusEmitter> = match args.format {
+        Format::JSON => Box::new(status_emitter::JSON),
+        Format::Pretty => Box::new(status_emitter::Text::verbose()),
+        Format::Terse => Box::new(status_emitter::Text::quiet()),
+    };
```
oli-obk, line 104: Add a method to `Format` or a free function to `status_emitter` to do this, and use it in all the duplicated places

tests/integrations/json-basic/tests/ui_tests.rs
```diff
@@ -0,0 +1,45 @@
...
+fn main() -> ui_test::color_eyre::Result<()> {
+    let path = "../../../target";
+    let mut config = Config {
+        output_conflict_handling: if std::env::var_os("BLESS").is_some() {
+            ui_test::bless_output_files
+        } else {
+            ui_test::error_on_output_conflict
+        },
+        bless_command: Some("cargo test".to_string()),
+        ..Config::rustc("tests/actual_tests")
```
oli-obk, line 12: instead of copying the basic test suite, use a path here pointing to the already existing test suite

tests/integrations/json-basic-fail/tests/ui_tests.rs
```diff
@@ -0,0 +1,35 @@
+use ui_test::{dependencies::DependencyBuilder, *};
+
+fn main() -> ui_test::color_eyre::Result<()> {
+    let path = "../../../target";
+    let mut config = Config {
+        // Never bless integrations-fail tests, we want to see stderr mismatches
+        output_conflict_handling: ui_test::error_on_output_conflict,
+        bless_command: Some("DO NOT BLESS. These are meant to fail".to_string()),
+        ..Config::rustc("tests/actual_tests")
```
oli-obk, line 9: same here

src/status_emitter/libtest_json.rs
oli-obk: Oh we can still emit it, just without the number of tests field

src/status_emitter/json.rs
oli-obk: oh nevermind 🤦 that's put into json

src/status_emitter/json.rs
```diff
@@ -0,0 +1,157 @@
...
+    path: PathBuf,
+    revision: String,
+    style: RevisionStyle,
+}
+
+impl TestStatus for JSONStatus {
+    /// A test has finished, handle the result immediately.
+    fn done(&self, result: &TestResult, aborted: bool) {
+        let status = if aborted {
+            String::from("timeout")
```
oli-obk, line 97:
```suggestion
            "timeout"
```

similarly below for the other status strings

src/status_emitter/json.rs
```diff
@@ -0,0 +1,157 @@
...
+// MAINTENANCE REGION START
+
+// When integrating with a new libtest version, update all emit_xxx functions.
+
+fn emit_suite_end(failed: usize, filtered_out: usize, ignored: usize, passed: usize, status: &String,) {
+    // Adapted from test::formatters::json::write_run_finish().
+    println!(r#"{{ "type": "suite", "event": "{status}", "passed": {passed}, "failed": {failed}, "ignored": {ignored}, "measured": 0, "filtered_out": {filtered_out} }}"#);
+}
+
+fn emit_test_end(name: &String, status: &String, error_output: &String) {
```
oli-obk, line 23:
```suggestion
fn emit_test_end(name: &str, status: &str, error_output: &str) {
```

(clippy CI will likely tell you that anyway ^^)

oli-obk:
please run rustfmt and clippy (`cargo clippy --fix` should work for most changes)

I think the reason you see some output twice is that e.g. for `//@run` tests you get one message for the test being compiled and one message for the compiled artifact being executed.

src/status_emitter/json.rs
```diff
@@ -0,0 +1,157 @@
...
+        let error_output = if let Err(errored) = result {
+            let command = &errored.command;
+            let stderr = errored.stderr.to_str_lossy();
+            let stdout = errored.stdout.to_str_lossy();
+            format!(r#"---- command\n\n{command}\n\n\---- stdout\n\n\{stdout}\n\n---- stderr\n\n\{stderr}"#)
+        } else {
+            String::new()
+        };
+
+        emit_test_end(&self.name, &status, &error_output);
```
oli-obk, line 114: the revision and the path are not part of the json, so you are getting duplicate messages. I'm ok with putting all of this into the name like `format!("{name}({revision}): {path}")` or sth. Usually when the revision is an empty string we don't print it at all (like not even empty parentheses). 

My preferred solution would be extra fields in the json, but I understand that this is suboptimal due to none of the json tooling consuming revision and path information yet.

oli-obk (changes requested):

oli-obk:
> I struggled with testing and ended up copying and adapting `integrations/basic` and `integrations/basic-fail` to work with the JSON emitter. I noticed a couple of issues:

bonus points if you don't need any duplication but can make it work by adding `#@revisions default json` to the `Cargo.toml`. It's gonna be icky telling the test about json, probably needs sth like `#@[json] rustc-env: USE_JSON=1` or going via compile-flags with like `-- --emit=json`

src/status_emitter/libtest_json.rs
kirtchev-adacore: OK, will do!

tests/integrations/json-basic/tests/ui_tests.rs
kirtchev-adacore: How did I not think of this ... will do!

src/status_emitter/json.rs
kirtchev-adacore: Note that I treat the path of the test as the name of the test. In addition, when creating a `TestStatus`, the revision is an empty string.

Is there a better way of obtaining the name and revision of a test?

oli-obk: You are already tracking it correctly in `for_revision and `for_path. So I think you just need to print it as part of the name

oli-obk:
> bonus points if you don't need any duplication but can make it work by adding `#@revisions default json` to the `Cargo.toml`. It's gonna be icky

This is just a random idea that I don't know if it will work. Just adjusting the path but duplicating the test is fine by me

src/status_emitter/json.rs
kirtchev-adacore: Fixed.

tests/integration.rs
kirtchev-adacore: I first tried a `Format` implementation, but both the use in `integrations.rs` and `lib.rs` did not like `status_emitter::from(args.format)`. I suspect that the path I used was missing something, but my Rust-fu is not strong enough.

A free function did the job.

tests/integrations/json-basic-fail/tests/ui_tests.rs
kirtchev-adacore: Same as above.

src/status_emitter/json.rs
kirtchev-adacore: Done.

tests/integrations/json-basic/tests/ui_tests.rs
oli-obk: That may just be my whacky test setup. So if there is too much trouble I'll check out your PR locally and fix it myself.

kirtchev-adacore: I wasn't able to figure out what the source of the auxiliary file build failure is, so I could use your help Oli.

I also noticed something else - the diffing portion of the testing seems to have a character limit. This was cutting off the terminating JSON brace among other things, and I thought that there was a general issue with the escaping. Here is what I observed:

Diff report:
image

Same "pure" output:
image

How do you want me to handle this?

kirtchev-adacore:
Oli, I think the PR is in good shape. There are a couple of outstanding issues which break testing:
* The diffing is slicing off some of the actual output of two tests from the `basic-fail` tests.
* The auxiliary file build failure in `basic` is registered as a failure even though the outputs satisfy the oracles.

oli-obk:
> * The diffing is slicing off some of the actual output of two tests from the `basic-fail` tests.

I think we can ignore that. `cargo test -- -- --bless` produces the "right" (matching) diff output, so while the diff output is broken, we can just keep it as is and I'll worry about it on the main branch.

> The auxiliary file build failure in basic is registered as a failure even though the outputs satisfy the oracles.

yea there was some weird stuff going on. I just merged the json tests into the regular test suite and that seems to have done the trick

oli-obk:
Thanks for bearing with me through all the change requests! I wanted a json backend for a while.

On that note, I assume bogus (unknown to you) json messages or extraneous json fields don't matter to you, right? So I can just add stuff even if you are an active consumer that will suddenly see more json than they need?

PR #319, opened 2025-05-05
Fix out-of-bounds error on multi-revision tests with error below markers

Jarcho:


tests/integrations/basic/tests/actual_tests/error_below_multi_rev.rs
```diff
@@ -0,0 +1,20 @@
...
+use ::std::vec;
+
+mod m {
+  //~[edition2015]v dead_code
+  pub fn foo() {}
+}
+
+//~[edition2015]| unused_imports
+//~[edition2018]v E0432
+use ::m::foo;
```
oli-obk, line 18: what was causing the panic here before?

src/parser.rs
```diff
@@ -382,11 +382,10 @@ impl CommentParser<Comments> {
                     }
 
                     for (span, line, idx) in delayed_fallthrough.drain(..) {
-                        if let Some(rev) = self
-                            .comments
-                            .revisioned
-                            .values_mut()
-                            .find(|rev| rev.error_matches[idx].line == line)
```
Jarcho, line 389: This was the cause. Only the revision each comment was added to was guaranteed to have the index.

oli-obk: oh right

oli-obk (approved):

PR #321, opened 2025-05-07
fix stale lockfiles

RalfJung:
A bunch of the lockfiles in this repo seem stale: just running `cargo check` changes their contents. This means just opening the repo in vscode introduces differences which is painful.^^

I'd expect this to fail CI but maybe there are `--locked` flags missing somewhere?


oli-obk:
> I'd expect this to fail CI but maybe there are `--locked` flags missing somewhere?

yea

Issue #323, opened 2025-05-07
GHA emitter panics when span points outside of a line


DaniPopes:
If an error is emitted where the span points to 1 character after the end of the line, the GHA emitter will panic. For example (a `ui_test` internal error):

```rust
@
//~^ ERROR:
```

with a normal emitter:
```
error: no pattern specified
 --> ...
  |
9 |     //~^ ERROR:
  |                ^
  |
```

with GHA emitter:
```
thread 'main' panicked at .../ui_test/src/status_emitter/gha.rs:20:14:
called `Option::unwrap()` on a `None` value
```

relevant code: https://github.com/oli-obk/ui_test/blob/a846e59cc71904d995edd98d4185476d6fb156c5/src/status_emitter/gha.rs#L17-L20

jenpaff:
@oli-obk could you give a rough estimate when we can expect a new release including the fix?

oli-obk:
Next week. I tried 3 times in the last 4 and got distracted every time 😅

oli-obk:
0.30 has been released with this fix

PR #325, opened 2025-05-07
Fix GHA emitter line count panic

DaniPopes:
Fixes https://github.com/oli-obk/ui_test/issues/323.


oli-obk:
Please add a test. It'll only get checked in CI, but that's ok

DaniPopes:
Where can I do that? integrations/basic-fail? I'm not sure if Gha would run on that

oli-obk:
Ah right they only run on top level things... I guess I need to figure out how to write gha tests.

Issue #326, opened 2025-05-26
JSON emitter fails to escape backslashes on Windows


kirtchev-adacore:
Hi Oli,

Here we go again! 😆 I am currently integrating `ui_test` `0.30.0` into our system, and `serde_json` fails to properly parse paths with `\` on Windows because the backslashes need to be escaped as `\\`:
```
Caused by:
    0: failed to parse message at line 15
    1: failed to parse message from slice '"{ \"type\": \"test\", \"event\": \"started\", \"name\": \"tests\\ui\\absurd-extreme-comparisons.rs () - tests\\ui\\absurd-extreme-comparisons.rs\" }"'
    2: invalid escape at line 1 column 58
```
The original JSON entry being parsed is actually
```
{ "type": "test", "event": "started", "name": "tests\ui\absurd-extreme-comparisons.rs () - tests\ui\absurd-extreme-comparisons.rs" }
```
and column 58 points at `\a`.

I looked at `PathBuf` and on the web how to deal with this. Folks are suggesting the use of the `Debug` implementation. Is this really the best option?

oli-obk:
There is `path.display().to_string()` that you can use and then escape the result of that. With debug you do get backslashes escaped, but nothing else, so I think we still need full json eacaping

PR #328, opened 2025-05-27
JSON emitter: escape backslashes on Windows

kirtchev-adacore:
This PR updates the JSON emitter to properly escape backslashes, which usually occurr on Windows, in test paths and revisions.

The PR also changes all `emit_xxx` function into `xxx_event` function which now construct a `String` which represents the related event, and adds unit tests to validate the `xxx_event` functions.

Fixes: https://github.com/oli-obk/ui_test/issues/326

* [ ] Add tests
* [ ] Add CHANGELOG.md entry

### DIFF
--- a/src/status_emitter/json.rs
+++ b/src/status_emitter/json.rs
@@ -13,45 +13,59 @@ use bstr::ByteSlice;
 
 // MAINTENANCE REGION START
 
-// When integrating with a new libtest version, update all emit_xxx functions.
+// When integrating with a new libtest version, update all xxx_event functions.
 
-fn emit_suite_end(failed: usize, filtered_out: usize, ignored: usize, passed: usize, status: &str) {
+fn suite_end_event(
+    failed: usize,
+    filtered_out: usize,
+    ignored: usize,
+    passed: usize,
+    status: &str,
+) -> String {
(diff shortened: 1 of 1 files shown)


kirtchev-adacore:
Oli, I ended up changing the `emit` functions into functions which return a `String` that I can validate in units tests. Since I did not want to make the new `xxx_event` functions public, I added the unit tests directly in `json.rs`.

src/status_emitter/json.rs
```diff
@@ -13,54 +13,68 @@ use bstr::ByteSlice;
...
-        r#"{{ "type": "test", "event": "started", "name": "{name} ({revision}) - {displayed_path}" }}"#
-    );
+    format!(
+        r#"{{ "type": "test", "event": "started", "name": "{escaped_name} ({escaped_revision}) - {escaped_path}" }}"#
+    )
 }
 
 // MAINTENANCE REGION END
 
+fn escape_backslashes(s: &str) -> String {
```
oli-obk, line 66: Is that correct? Don't you need to escape everything?

kirtchev-adacore: Done. The code actually ended up looking way better!

kirtchev-adacore:
Fantastic, thanks Oli!

Issue #327, opened 2025-05-26
does not support crates that appear as both build-dependencies and core dependencies


RalfJung:
Currently, `ui_test` doesn't support crates that appear as both build-dependencies and core dependencies. That's a bummer as it blocks https://github.com/rust-lang/rust-clippy/pull/14883 which is on the critical path to resolving https://github.com/rust-lang/rust/issues/78717.

So, I wonder what could be done here. First I tried to understand what happens.

### The problem

To determine the rlib files that should be added to the tests, ui_test builds the test dependency crate with `--message-format=json`. Then it scours the JSON we get from that for the files we care about.

The problem is that sometimes we get this:
```
{"reason":"compiler-artifact","package_id":"registry+https://github.com/rust-lang/crates.io-index#quote@1.0.40","manifest_path":"/home/r/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/quote-1.0.40/Cargo.toml","target":{"kind":["lib"],"crate_types":["lib"],"name":"quote","src_path":"/home/r/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/quote-1.0.40/src/lib.rs","edition":"2018","doc":true,"doctest":true,"test":true},"profile":{"opt_level":"0","debuginfo":2,"debug_assertions":true,"overflow_checks":true,"test":false},"features":["default","proc-macro"],"filenames":["/home/r/src/rust/clippy/target/debug/deps/libquote-d5accd6216ede55c.rlib","/home/r/src/rust/clippy/target/debug/deps/libquote-d5accd6216ede55c.rmeta"],"executable":null,"fresh":false}
[...]

PR #329, opened 2025-06-30
dependencies: heuristic to distinguish host and target artifacts

RalfJung:
Fixes https://github.com/oli-obk/ui_test/issues/327 by looking at the path in which artifacts are stored: if we pass `--target` to cargo, that will be different for host and target crates. If a non-proc-macro host crate shows up, we just ignore it during artifact gathering.

I have confirmed that in the context of https://github.com/rust-lang/rust-clippy/pull/14883 this makes the dependency build work.

* [x] Add tests
* [x] Add CHANGELOG.md entry
* [x] Bumped minor version and committed lockfiles in case a release is desired

### DIFF
--- a/tests/integrations/dep-fail/Cargo.lock
+++ b/tests/integrations/dep-fail/Cargo.lock
@@ -461,7 +461,9 @@ dependencies = [
 
 [[package]]
 name = "spanned"
-version = "0.4.0"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c92d4b0c055fde758f086eb4a6e73410247df8a3837fd606d2caeeaf72aa566d"
 dependencies = [
  "anyhow",
  "bstr",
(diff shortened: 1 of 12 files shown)


RalfJung:
> Add CHANGELOG.md entry

Should this go into the changelog?

oli-obk (approved):

oli-obk:
You new ci checks for lockfiles seem to be working

RalfJung:
Oddly, a bunch of lockfiles did update when I just changed the version, but the main one did not.^^

RalfJung:
Nothing ensures that the tests have updated lockfiles though... ah well, another time.

tests/integrations/dep-fail/Cargo.lock
RalfJung, line 1: This lockfile and ui_test_dep_bug also have some more changes from the "spanned" thing; doesn't seem worth a separate PR.

PR #332, opened 2025-07-01
rust-toolchain: add rust-src component

RalfJung:
rust-src is pretty small, and needed for RA. If we include clippy by default, surely we should also include this.

Also convert the file to Unix line endings. It used Windows line endings for some reason...


oli-obk:
Probably because I created it on windows and git cross platform is a mess

PR #338, opened 2025-07-19
bump colored and MSRV

RalfJung:
Bumping colored removes some old windows crate from the dependencies, and also removes one of the many dependencies on lazy_static.

Colored 3.0 has an MSRV of 1.80, which is almost a year old so I hope that's fine.


RalfJung:
Hm, the new rustc needs some ref file changes but I can't figure out how to bless these tests. The contributor guide says
```
Running `cargo test` will automatically update the `.stderr`
and `.stdout` files.
```
but that's obviously not true, `cargo test` fails...

oli-obk:
Oh I thought the negative diff was "file deleted" 🤦

PR #343, opened 2025-11-01
text emitter: use singular form when there was only 1 unmatched diagnostic

ada4a:
* [x] Add tests
    * .. but wasn't able to bless them as the test suite seems to hang for some reason
* [x] Add CHANGELOG.md entry

### DIFF
--- a/tests/integrations/basic-fail/Cargo.stdout
+++ b/tests/integrations/basic-fail/Cargo.stdout
@@ -5,6 +5,7 @@ test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; fini
 
 { "type": "suite", "event": "started" }
 { "type": "test", "event": "started", "name": "tests/actual_tests/bad_pattern.rs () - tests/actual_tests/bad_pattern.rs" }
+{ "type": "test", "event": "started", "name": "tests/actual_tests/bad_pattern_multiple.rs () - tests/actual_tests/bad_pattern_multiple.rs" }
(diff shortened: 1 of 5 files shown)


oli-obk:
hmm I'll have to try the test suite locally, but are you sure it hangs and doesn't just take forever (it rebuilds a lot of things on the first run)

ada4a:
That does sound right -- many of the changes seem to come from improved rustc diagnostics

EDIT: Ahh, right! That's because I have Cargo installed as a free-standing program (so not from rustup), which is why `rust-toolchain.toml` isn't respected and I end up running the latest stable instead of 1.80.1. Apologies for the confusion

oli-obk:
Great! Thanks for the fix

oli-obk:
Hmm I saw the test changes for 1 unmatched diagnostic and forgot to check that there's still a case for 2+

tests/integrations/basic-fail/Cargo.stdout
```diff
@@ -132,6 +135,135 @@ full stdout:
...
+ --> tests/actual_tests/bad_pattern_multiple.rs:7:17
+  |
+7 |     //~^ ERROR: miesmätsched types
+  |                 ^^^^^^^^^^^^^^^^^^ expected because of this pattern
+  |
+
+error: there was 1 unmatched diagnostic
+ --> tests/actual_tests/bad_pattern_multiple.rs:4:9
+  |
+4 |     add("42", 3);
```
oli-obk: Maybe there need to be multiple per line for there to be just one diagnostic informing us about multiple cases

PR #347, opened 2025-11-14
Bump some dependencies so owo-colors 3.x is gone

RalfJung:
There's apparently no 3.x release with https://github.com/owo-colors/owo-colors/pull/131, but 4.x has it, so this is the easiest way to get rid of old owo-colors without that PR.

That should help with https://github.com/rust-lang/rust/pull/148952.


oli-obk:
I can do that if you do the version bump and lockfile update that cargo test does

PR #368, opened 2026-05-04
New cargo layout

ranger-ross:
This PR adds support for Cargo's new `build-dir` layout tracked in https://github.com/rust-lang/cargo/issues/15010.
I opted to not modify the tests as some of them have paths and it felt like it was going to be painful to keep both layouts in tests.
I think we can update the tests when the new `build-dir` lands on stable. wdyt?

Also I'd appreciate if you could cut a new release once this is merged as it blocks https://github.com/rust-lang/rust/pull/155439

* [ ] Add tests
* [x] Add CHANGELOG.md entry
* [x] Bumped minor version and committed lockfiles in case a release is desired

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ui_test"
-version = "0.30.4"
+version = "0.30.5"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 description = "A test framework for testing rustc diagnostics output"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+### Changed
+
+## [0.30.5] - 2026-05-04
+
+### Added
+
+- Added support for Cargo's new `build-dir` layout
+
+### Fixed
+
 - Update the help string in CLI to include new flags
 
 ### Changed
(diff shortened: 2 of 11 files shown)


oli-obk:
> I think we can update the tests when the new `build-dir` lands on stable. wdyt?

yea that wfm

is there a better way to detect whether the new layout is active? Some env var or -Z flag that we could process and switch?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ui_test"
-version = "0.30.4"
+version = "0.31.0"
```
oli-obk: this can just be a patch version (which for 0.x crates is a minor version)

ranger-ross:
> is there a better way to detect whether the new layout is active? Some env var or -Z flag that we could process and switch?

There is `-Zbuild-dir-new-layout` (and `CARGO_UNSTABLE_BUILD_DIR_NEW_LAYOUT`).
Though the goal is to stabilize this soon. (The [stabilization PR](https://github.com/rust-lang/cargo/pull/16807) has already been FCP'd and we just need to enable it in rl/r so that we can update the Cargo submodule)

Cargo.toml
ranger-ross: sure, changed this to `0.30.5`

ranger-ross:
btw, thank you for the quick review and release! :heart:

PR #369, opened 2026-05-11
bump colored, ctrlc

RalfJung:
This removes windows-targets from the lockfile.

Also `cargo update` all the integration tests.


oli-obk:
This crate is a library. I think that means users' code breaks if we use logic from deps that isn't required by Cargo.toml

RalfJung:
Yeah I also realized that I didn't end up having to change the Cargo.toml here.

But it still seems worth getting windows-targets out of this repo, no?

oli-obk:
> Yeah I also realized that I didn't end up having to change the Cargo.toml here.

uh. I meant the opposite. We need to change the `Cargo.toml` otherwise we're accidentally gonna cause some breakage in the future when relying on new `colored` APIs or sth

RalfJung:
Ah. Yeah I guess ideally we'd have a CI job that tests with "minimal version" resolution but the Rust ecosystem is generally pretty bad at that (mostly because cargo has no stable support for it).

I bumped them in the toml file.

PR #371, opened 2026-05-12
set spinner to print to stdout

RalfJung:
Everything else we print seems to be printed to stdout, but the spinner defaults to stderr according to [indicatif docs](https://docs.rs/indicatif/latest/indicatif/struct.ProgressBar.html#method.new_spinner).

### DIFF
--- a/src/status_emitter/text.rs
+++ b/src/status_emitter/text.rs
@@ -98,13 +98,13 @@ impl Text {
         let (sender, receiver) = crossbeam_channel::unbounded();
         #[cfg(feature = "indicatif")]
         let handle = std::thread::spawn(move || {
-            let bars = MultiProgress::new();
-            let progress = match progress {
-                OutputVerbosity::Progress => bars.add(ProgressBar::new(0)),
-                OutputVerbosity::DiffOnly | OutputVerbosity::Full => {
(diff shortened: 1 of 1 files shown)


oli-obk:
I never realized this was relevant because it was all a TUI. Have you tested that it still works as expected and we're not getting any issues with buffers?

RalfJung:
If I run `./miri test 2>/dev/null` it prints some stuff but not the live progress. If I run `./miri test >/dev/null` it prints just the live progress.

But, somehow it still does the same with this PR applied?!?

RalfJung:
And yeah something breaks with this PR, the progress bar moves to the top of the terminal. I have no idea what's happening. Even with a TUI, it has to send the magic characters on either stdout or stderr, doesn't it?

RalfJung:
Actually... the main reason I was surprised about the progress bar at the top is that currently one sees *no* progress bar with `./miri test`. I could have sworn there used to be a progress bar... but it seems at some point it disappeared? And this PR makes it reappear.

But please try this our yourself, I am very confused right now.^^

oli-obk:
uh, I can confirm. The progress bar has disappeared at some point, but with your changes it's back 🫠

oli-obk:
Ok, let's just merge this for now while I'm evaluating options. I have considered moving away from indicatif and thus `console` and try out `ratatui`

PR #372, opened 2026-05-29
make Errored fields public

RalfJung:
This is needed to implement something like `Run` as a user-defined flag.

Also, clarify the polarity of `test_condition`.

* [ ] Add tests -> does not seem needed?
* [x] Add CHANGELOG.md entry


RalfJung:
Should I make a PR for a release?

oli-obk:
I'll do that locally, but yea, if you want sth immediately, just add the version bumps to the same PR

oli-obk:
released

RalfJung:
Thanks :)

PR #373, opened 2026-06-02
cleanup: update to more recent cargo{_metadata,-platform}

durin42:
I noticed this because our CVE-bot at work complained about our old cargo_platform version kicking around in service of ui_test (via clippy.) I figured it made sense to clean things up while I'm thinking about it.

* [ ] Add tests
* [ ] Add CHANGELOG.md entry

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,34 +88,34 @@ checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
 
 [[package]]
 name = "camino"
-version = "1.1.9"
+version = "1.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


durin42:
FWIW a release containing this would be nice, but it's probably not urgent or anything.

oli-obk:
Rustfmt and clippy are failing.

Happy to do a quick release. If you bump the patch version in this PR, too, I can do that quickly. You'll probably need to bump the dep in clippy and miri for it to make a difference for you

durin42:
Bumped version, fixed clippy/fmt issues (I hope on those last bits - should be good tho)

Thanks!

oli-obk:
looks like you need to rerun tests after the version bump so that the lockfiles get updated

durin42:
Whoops, done. Didn't expect that (and as you probably guessed, bumping the version was the last step I performed.)

oli-obk:
Thanks! New version should be published in a few mins

durin42:
Awesome thanks!

PR #374, opened 2026-06-25
DependencyBuilder: fix bless_lockfile behavior

RalfJung:
The dependency builder currently always blesses the lockfile, no matter what `bless_lockfile` says: the first invocation (that does the actual build) is never done with `--locked`. The 2nd invocation (for the metadata) uses `--locked` but at that point the lockfile has already been updated.

Only the first invocation needs the flag (if the lockfile is outdated it will fail).

### DIFF
--- a/src/dependencies.rs
+++ b/src/dependencies.rs
@@ -87,12 +87,9 @@ fn build_dependencies_inner(
         }
     }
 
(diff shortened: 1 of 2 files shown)


oli-obk (approved):
Oh wow... Added the closure and didn't call it where its logic was called before

